PR #6, opened 2017-08-13
KML Integration

shivvo:
I've added a model that takes in a placemark KML string and does the proper parsing. I thought it would be better to integrate it into the model's constructor for ease of use, but the code can be easily factored out in case that is not preferred.

### DIFF
--- a/src/models/Kml.js
+++ b/src/models/Kml.js
@@ -1,16 +1,70 @@
 
 import Location from './Location';
+import Stop from './Stop';
+import KmlUtils from '../utils/KmlUtils'
 
 class Kml {
+  // Line number of the bus this KML corresponds to
   lineNumber: number;
+  // The full placemark - in case we need it in future revisions of the implementation
   placemark: string;
+  // The placemark "prefix" - the LEADING portion of this KML that does not contain coordinates
+  placemarkPrefix: string;
+  // The placemark "suffix" - the TRAILING tion of this KML that does not contain coordinates
+  placemarkSuffix: string;
+  // The list of all locations in the KML
   locations: Array<Location>;
 
-  constructor (lineNumber: number, placemark: string, locations: Array<Location>) {
+  constructor (lineNumber: number, placemark: string) {
     this.lineNumber = lineNumber;
     this.placemark = placemark;
-    this.locations = locations; 
+
+    // The format is longitude,latitude,0 (why 0? TCAT doin a bamboozle?)
+    var regex = /-?[\d|.|e|E|\+]+,-?[\d|.|e|E|\+]+,-?[\d|.|e|E|\+]+/g;
+    var locations = placemark
+      .match(regex);
+
+    // first and last locations in the KML
+    var firstLocation = locations[0];
+    var lastLocation = locations[locations.length-1];
+
+    // The indices to use when slicing the KML into the prefix and suffix
+    var locationsStart = placemark.indexOf(firstLocation);
+    var locationsEnd = placemark.indexOf(lastLocation) + lastLocation.length;
+
+    this.placemarkPrefix = placemark.substring(0, locationsStart);
+    this.placemarkSuffix = placemark.substring(locationsEnd);
+    this.locations = locations
+      .map(a => a
+        .split(",") // split into three number strings
+        .map(b => Number(b)) // convert the numbers
+        .slice(0,2)) // Grab the first two (third provided but irrelevant)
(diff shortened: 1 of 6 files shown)


src/utils/KmlUtils.js
```diff
@@ -0,0 +1,41 @@
+
```
Jma353: It looks like this file, uses either 4-spaces or tabs to indent, can you convert to 2 spaces?

src/utils/KmlUtils.js
```diff
@@ -0,0 +1,41 @@
+
+import Location from '../models/Location';
+
+const locationsFromPlacemark = (placemark: string): Array<Location> => {
+	var regex = /-?[\d|.|e|E|\+]+,-?[\d|.|e|E|\+]+,-?[\d|.|e|E|\+]+/g;
+	var start = 0;
+	for (let i = 0; i < placemark.length; i++) {
+		var jspace = placemark.indexOf(" ", i);
+		var substr = placemark.substring(i, jspace);
+		if (substr.match(regex) != null) {
```
Jma353: `!==` rather than `!=`

src/utils/KmlUtils.js
```diff
@@ -0,0 +1,41 @@
...
+			start = i;
+			break;
+		}
+	}
+
+	var end = 0;
+	for (let i = start; i < placemark.length; i++) {
+		var jspace = placemark.indexOf(" ", i);
+		var substr = placemark.substring(i, jspace);
+		if (substr.match(regex) == null) {
```
Jma353: Ditto, `===` rather than `==`

src/utils/KmlUtils.js
```diff
@@ -0,0 +1,41 @@
...
+			end = i;
+			break;
+		}
+		i += substr.length;
+	}
+
+	var coordinateStr = placemark.substring(start, end);
+	var locations = coordinateStr
```
Jma353: Add light comments here to indicate what you're doing in this series of maps / splits / slices, so that it's readable if we ever need to revisit this code.  I understand the data-processing you're doing, but it takes me a few seconds to wrap my head around the nested maps and such.

Also, if you're performing multi-line, nested functionality in a `map` function, it makes it easier to read if you do the following:

````
// previous data processing
.trim() 
.split(" ")
.map(a: DECLARE_A_TYPE => {
  return a.split(",").map(b: DECLARE_A_TYPE => Number(b)).slice(0, 2);
})
// etc., you get the point
````

File: src/models/Location.js [hunk 5]
```diff
@@ -19,6 +19,13 @@ class Location {
     );
   }
 
+  distance (location: Location): number {
```
Jma353: Is there no built-in JS function for this?

src/utils/KmlUtils.js
```diff
@@ -13,6 +14,34 @@ const locationsFromPlacemark = (placemark: string): Array<Location> => {
...
+    if (kml.locations[i].distance(stop1.location) < kml.locations[i].distance(kml.locations[start])) {
+      start = i;
+    }
+
+    if (kml.locations[i].distance(stop2.location) < kml.locations[i].distance(kml.locations[end])) {
+      start = i;
+    }
+  }
+
+  function arrayRotate(arr, count) {
```
Jma353: Make this arrow-function.

src/utils/KmlUtils.js
```diff
@@ -13,6 +14,34 @@ const locationsFromPlacemark = (placemark: string): Array<Location> => {
   return locations;
 }
 
+
+
+const kmlApproximation = (kml, stop1: Stop, stop2: Stop): Array<Location> => {
```
Jma353: Comment this function slightly so it's easier to understand.

src/utils/KmlUtils.js
Jma353: Also give a one-line comment on top of or in this function to declare what it's purpose is.  `arrayRotate` is kinda vague.

src/utils/KmlUtils.js
```diff
@@ -0,0 +1,14 @@
+
```
Jma353: `// @flow` at the top so there's type-checking.

src/models/Kml.js
```diff
@@ -0,0 +1,70 @@
+
```
Jma353: `// @flow`

src/models/Location.js
Jma353: Is this supposed to be Euclidean distance? If so, square-root.

Jma353:
@shivvo done reviewing, feel free to fix the things I mentioned and merge this in.

PR #9, opened 2017-09-02
Austin/live tracking

AAAstorga:
Added a live tracking endpoint `/tracking/`
Tested and it works and returns all the data requested by frontend

### DIFF
--- a/src/routers/GetBusTrackingRouter.js
+++ b/src/routers/GetBusTrackingRouter.js
@@ -49,7 +49,7 @@ class GetBusTrackingRouter extends AppDevRouter {
       });
       return parsedData;
     } catch (error) {
-      // @todo log error on backend but pass in a presentable string so frontend can display it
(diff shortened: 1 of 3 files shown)


src/routers/GetBusTrackingRouter.js
```diff
@@ -0,0 +1,57 @@
...
+          opStatus: busInfo.OpStatus,
+          routeID: busInfo.RouteId,
+          runID: busInfo.RunId,
+          speed: busInfo.Speed,
+          tripID: busInfo.TripId,
+          vehicleID: busInfo.VehicleId
+        };
+      });
+      return parsedData;
+    } catch (error) {
```
Jma353: Maybe throw a custom error message in the future?  Add a `// TODO` here so we can know to come back and reflect a proper error to client, as the message will eventually bubble up to them.

PR #11, opened 2017-09-06
Sanjana/endpoints

skaundinya15:
sample endpoint

I've attached the sample endpoint that I tested it with from the git issue. The way I constructed this as @Jma353 talked about on Sunday was that each route is bound by walking. So from the current location, the user walks, and from the end location the user walks. Further more, I thought it might be useful to redefine how the type place is used. So in my changes I made it such that the end destination is always going to be a place, even if the end destination is just a bus stop. I figured this might be easier to understand as otherwise it would be hard to distinguish if a user wants their end location to be a stop. As in the example endpoint in the git issue, the end location is a bus stop, but in my implementation I've treated it as a place so that it's indicative that the route has ended.

### DIFF
--- a/src/routers/GetRouteRouter.js
+++ b/src/routers/GetRouteRouter.js
@@ -79,16 +79,32 @@ class GetRouteRouter extends AppDevRouter {
     const arrivalTime = Math.floor(leaveBy + (endTime - startTime));
 
     // Then build up stop ordering / walking information
-    let mainStops = [];
-    let mainStopNums = [];
-    for (let i = 0; i < result.length; i++) {
+    const stopSummary = [];
+    for (let i = 0; i <result.length; i++) {
       const r = result[i];
-      if (r.endStop.name === TCATConstants.END_WALKING) {
-        mainStopNums.push(TCATConstants.WALKING_TCAT_NUMBER);
-      } else if (r.endStop.name !== TCATConstants.START_WALKING) {
-        mainStopNums.push(r.busNum);
-        mainStops.push(r.endStop.name);
+      const obj = {};
+      if (i == 0 && r.startStop.name == START_WALKING) {
+        obj.name = "Current Location";
+        obj.type = "currentLocation";
+        obj.busNumber = r.busNum;
+      } else {
+        obj.name = r.startStop.name;
+        obj.busNumber = r.busNum;
       }
+      if ((result[i+1].busNum == -1) && (i < result.length)) {
+          obj.nextDirection = "walk";
+      } else if ((result[i+1].busNum == null) && (i < result.length)) {
+          obj.nextDirection = "none";
+      } else {
+        obj.nextDirection = "bus";
+      }
+      if (obj.nextDirection == "bus" || obj.nextDirection == "walk") {
+        obj.type = "stop";
+      }
+      if ((i == result.length - 1) && (r.endStop.name == END_WALKING)) {
+        obj.type = "place";
+      }
+      stopSummary.append(obj);
     }
 
     const kmls = this._grabKMLsFromRoute(mainStops, mainStopNums);
@@ -100,11 +116,9 @@ class GetRouteRouter extends AppDevRouter {
       // Main data
       departureTime: departureTime,
       arrivalTime: arrivalTime,
-      mainStops: mainStops,
-      mainStopNums: mainStopNums,
+      stopSummary: stopSummary,
       kmls: kmls
     }];
   }
-}
 
 export default new GetRouteRouter().router;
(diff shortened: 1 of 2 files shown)


Jma353:
In the tested example, it labels "Seneca @ Commons" as a "place," but it's actually a "stop."

src/routers/GetRouteRouter.js
```diff
@@ -91,6 +92,37 @@ class GetRouteRouter extends AppDevRouter {
       }
     }
 
+     const stopSummary = [];
+     for (let i = 0; i <result.length; i++) {
+       const r = result[i];
+       const obj = {};
```
Jma353: Better variable name

src/routers/GetRouteRouter.js
```diff
@@ -91,6 +92,37 @@ class GetRouteRouter extends AppDevRouter {
       }
     }
 
+     const stopSummary = [];
```
Jma353: Overall comment: these different string should be constants.  "walk" should be a constant, "place" should be a constant, etc.  If we need to change these every, we'd have to go back through the code and change the use of every one of these plain-text strings.

src/routers/GetRouteRouter.js
```diff
@@ -91,6 +92,37 @@ class GetRouteRouter extends AppDevRouter {
...
+     for (let i = 0; i <result.length; i++) {
+       const r = result[i];
+       const obj = {};
+      if (i == 0 && r.startStop.name == "START_WALKING") {
```
Jma353: ===, not ==

src/routers/GetRouteRouter.js
```diff
@@ -79,6 +79,7 @@ class GetRouteRouter extends AppDevRouter {
     const arrivalTime = Math.floor(leaveBy + (endTime - startTime));
 
     // Then build up stop ordering / walking information
+
```
Jma353: Remove

src/routers/GetRouteRouter.js
```diff
@@ -91,6 +92,37 @@ class GetRouteRouter extends AppDevRouter {
       }
     }
 
+     const stopSummary = [];
+     for (let i = 0; i <result.length; i++) {
+       const r = result[i];
```
Jma353: Better variable name

src/routers/GetRouteRouter.js
```diff
@@ -91,6 +92,37 @@ class GetRouteRouter extends AppDevRouter {
       }
     }
 
+     const stopSummary = [];
+     for (let i = 0; i <result.length; i++) {
```
Jma353: `i < result.length`

src/routers/GetRouteRouter.js
```diff
@@ -91,6 +92,37 @@ class GetRouteRouter extends AppDevRouter {
...
+        obj.name = "Current Location";
+        obj.type = "currentLocation";
+        obj.nextDirection = "walk";
+        obj.busNumber = r.busNum;
+      } else {
+        obj.name = r.startStop.name;
+        obj.busNumber = r.busNum;
+      }
+      if ((i + 1 < result.length) && (obj.nextDirection == null)) {
+        if ((r.busNum != -1) && (r.busNum != null)) {
```
Jma353: `if (r.busNum && r.busNum !== -1) { ... }`

Jma353:
Overall comment - I was expecting you to use my `mainStops` / `mainStopNums` as a tool to format this data, as those two data structures are extremely organized and linear in their structure.  Right now this code has a lot of `if`-statement base conditions that are prone to edge-case scenarios.  I'd recommend refactoring to use the two arrays constructed.

skaundinya15:
Regarding the last stop being labeled as a place, I thought about it and realized that I couldn't find a way to distinguish between a place or a stop for the very end. In other words, if the last point is a stop, I can't tell that it's a stop as I only know the lat and long of that location. So I thought it might be useful for the frontend to label the last stop as a place as thats where the user will be getting off so even if they get off at a stop, it would just be labeled as a place because thats their end destination. I was just thinking this could be a way to do it but if you think there's another way I can distinguish between those two, let me know

skaundinya15:
screen shot 2017-09-07 at 1 28 25 am

Here's the updated screenshot with the final changes I have. @Jma353 I know we talked about using your data structures to do some post processing, so I used them in if conditions to check for certain conditions to return the array of JSON properly. I think these changes will work for the time being; in the future when we change the way we bound routes, we can revisit this to make changes accordingly. However for now I think it's good to get it shipped out now so we can meet our deadline.

PR #24, opened 2018-02-25
Hot Fixes

mattbarker016:
- Changed response for routing `data` from `JSON.stringify(possibleRoutes)`to just `possibleRoutes`.
- Added `*graph-cache*` to `.gitignore`


src/RouteRouter.js
```diff
@@ -141,8 +141,9 @@ class RouteRouter extends AbstractRouter {
                 })
 
             }
-            //res.set('Content-Type', 'application/json');
-            return JSON.stringify(possibleRoutes);
+            // res.set('Content-Type', 'application/json');
+            // return JSON.stringify(possibleRoutes);
+            return possibleRoutes
```
AAAstorga: If you don't want it stringified, please do it on all Routers. Thanks

AAAstorga: Also please include an explanation on why this change is happening

mattbarker016: The value for the data key was an array made into a string. You can return the array object, which makes it easier to iterate over the unwrapped data on the client-side.

AAAstorga:
Looks good. Please merge into unstable branch

PR #26, opened 2018-03-02
Arman/access key

arman-esmaili:
Fixes network requests, should support the remaining routes


src/TokenUtils.js
```diff
@@ -0,0 +1,51 @@
...
+import fs from 'fs';
+import https from 'https';
+import axios from 'axios';
+import qs from 'qs';
+import credentials from '../config.json';
+
+function isAccessTokenExpired() {
+	let currentDate = new Date();
+	let expiryDate = new Date(credentials.expiry_date);
+	return expiryDate.getTime() < currentDate.getTime();
```
AAAstorga: Also maybe do like a 30 second offset. Just incase you check like 1 second before it expires, then it says it is good to go, but when you make the request it fails

src/TrackingRouter.js
```diff
@@ -9,10 +10,14 @@ class TrackingRouter extends AbstractRouter {
     }
 
     async content(req: Request): Promise<any> {
-        let routeID = req.query.routeID;
-        const AuthStr = 'Bearer 5a54bc7f-a7df-3796-a83a-5bba7a8e31c8';
         try {
-        let trackingRequest = await axios.get('https://realtimetcatbus.availtec.com/InfoPoint/rest/Vehicles/GetAllVehiclesForRoute?routeID=' + routeID, {headers: {Authorization: AuthStr}});
+        let parameters: any = {
+        	routeID: req.query.routeID
+        };
```
AAAstorga: nice!

src/TrackingRouter.js
```diff
@@ -9,10 +10,14 @@ class TrackingRouter extends AbstractRouter {
...
+        let trackingRequest = await axios.get('https://realtimetcatbus.availtec.com/InfoPoint/rest/Vehicles/GetAllVehiclesForRoute', {
```
AAAstorga: This url should actually be from the official endpoint, like the AllBuStopsRouter. Change it to that and also use the accessToken

src/TokenUtils.js
arman-esmaili: a 30 second buffer is not effective b/c the API will return the same token until it officially expires. therefore, a 0.5 second may even be too long. something to consider further

PR #31, opened 2018-03-16
Austin/walking g hopper

AAAstorga:
-Added walking routes
-fixed tracking so it would only show relevant buses
- minor fixes for frontend

### DIFF
--- a/scripts/graph-walking.sh
+++ b/scripts/graph-walking.sh
@@ -1,6 +1,2 @@
 rm -rf graphhopper-walking/graph-cache
 java -Xmx1g -Xms1g -jar graphhopper-walking/graphhopper/web/target/graphhopper-web-*-with-dep.jar datareader.file=osrm/map.osm jetty.port=8987 jetty.resourcebase=./graphhopper/web/src/main/webapp graph.flag_encoders=foot prepare.ch.weightings=no graph.location=graphhopper-walking/graph-cache
-
-
-in node Dockerfile add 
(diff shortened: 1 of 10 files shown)


scripts/graph-walking.sh
```diff
@@ -0,0 +1,6 @@
+rm -rf graphhopper-walking/graph-cache
+java -Xmx1g -Xms1g -jar graphhopper-walking/graphhopper/web/target/graphhopper-web-*-with-dep.jar datareader.file=osrm/map.osm jetty.port=8987 jetty.resourcebase=./graphhopper/web/src/main/webapp graph.flag_encoders=foot prepare.ch.weightings=no graph.location=graphhopper-walking/graph-cache
+
+
+in node Dockerfile add 
+RUN git checkout austin/routingOptimizations
```
arman-esmaili: should this git directory be change to master for deployment?

src/WalkingUtils.js
```diff
@@ -0,0 +1,64 @@
...
+    };
+
+    let endCoords = {
+        lat: path.points.coordinates[path.points.coordinates.length - 1][1],
+        long: path.points.coordinates[path.points.coordinates.length - 1][0]
+    };
+
+    let numberOfTransfers = 0
+
+    let walkingPath = path.points.coordinates.map(point => {
```
arman-esmaili: does this path contain more than two coordinates for proper map matching? or do we need to add the average lat/long again?

Issue #42, opened 2018-09-05
Add automated endpoint tests


zeladada:
Automated endpoint/end-to-end/route tests are needed so we can quickly verify basic functionality during development. We can add more specific unit and integration tests later on.

The project is already configured to test with Jest and Supertest and a testing file is already present in the src>test folder with skeleton code for all routes. The tests will run automatically after a src file change when running in dev mode.

zeladada:
Added skeleton code, basic endpoint functionality tests, and more with #43. @JunyoungLim will add some more tests while working on #46 to ensure the changes do not break routing.

zeladada:
#70 done

Issue #38, opened 2018-08-27
Unhandled promise rejection warning on npm start


zeladada:
![image](https://user-images.githubusercontent.com/7935434/44638224-b9ee1d00-a983-11e8-9534-02787ee79343.png)

zeladada:
![image](https://user-images.githubusercontent.com/7935434/45446156-282c2280-b69a-11e8-8f87-bfb8e70b5575.png)

zeladada:
![image](https://user-images.githubusercontent.com/7935434/45446788-c53b8b00-b69b-11e8-85eb-f65b7014610b.png)

Run `flow` to see type issues, then resolve and see if the problem persists

PR #58, opened 2018-09-16
Nate/fix flow errors

zeladada:
Fixes some existing flow errors, including one that would cause the delay route to always throw an undefined variable error. #38 Better tests needed!! #42

### DIFF
--- a/src/routers/TrackingRouter.js
+++ b/src/routers/TrackingRouter.js
@@ -19,7 +19,7 @@ class TrackingRouter extends AppDevRouter<Object> {
     async content(req: Request): Promise<any> {
         const trackingArray = req.body.data;
         let foundTrackingData = false;
-        let trackingInformation = [];
+        const trackingInformation = [];
         let noData = false;
 
         for (let index = 0; index < trackingArray.length; index++) {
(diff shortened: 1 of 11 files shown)


src/routers/TrackingRouter.js
```diff
@@ -45,49 +45,49 @@ class TrackingRouter extends AppDevRouter<Object> {
...
+                                latitude: busInfo.Latitude,
+                                longitude: busInfo.Longitude,
+                                name: busInfo.Name,
+                                opStatus: busInfo.OpStatus,
+                                routeID: busInfo.RouteId,
+                                runID: busInfo.RunId,
+                                speed: busInfo.Speed,
+                                tripID: busInfo.TripId,
+                                vehicleID: busInfo.VehicleId,
+                                case: 'validData',
```
young-k: any reason why everything is alphabetized except `case`?

src/utils/AlertsUtils.js
```diff
@@ -21,29 +21,31 @@ async function fetchAlerts() {
...
+                message: alert.Message,
+                fromDate: parseMicrosoftFormatJSONDate(alert.FromDate),
+                toDate: parseMicrosoftFormatJSONDate(alert.ToDate),
+                fromTime: parseMicrosoftFormatJSONDate(alert.FromTime),
+                toTime: parseMicrosoftFormatJSONDate(alert.ToTime),
+                priority: alert.Priority,
+                daysOfWeek: getWeekdayString(alert.DaysOfWeek),
+                routes: alert.Routes,
+                signs: alert.Signs,
+                channelMessages: alert.ChannelMessages,
```
young-k: similar comment here, can this be alphabetized?

src/utils/AllStopUtils.js
```diff
@@ -22,21 +22,23 @@ async function fetchAllStops() {
...
-        allStops = stopsRequest.map(stop => ({
-            name: stop.Name,
-            lat: stop.Latitude,
-            long: stop.Longitude,
-        }));
+        if (stopsRequest) {
+            allStops = JSON.parse(stopsRequest).map(stop => ({
+                name: stop.Name,
+                lat: stop.Latitude,
+                long: stop.Longitude,
```
young-k: same thing

src/routers/TrackingRouter.js
```diff
@@ -45,49 +45,49 @@ class TrackingRouter extends AppDevRouter<Object> {
...
-                            tripID: busInfo.TripId,
-                            vehicleID: busInfo.VehicleId,
-                            case: 'validData',
-                        };
                     });
 
+                    const trackingData = trackingRequest
+                        && JSON.parse(trackingRequest).data.filter(
+                            busInfo => busInfo.VehicleId === realtimeData.vehicleID,
+                        ).map((busInfo) => {
```
young-k: Before you merge, can you make an issue to refactor this? It's pretty bulky & tough to understand

src/utils/AlertsUtils.js
zeladada: I think alphabetizing keys adds a lot of needless complexity and is difficult to maintain. I can make an ESlint rule to alphabetize everything automatically if you want, but I also don't think alphabetizing is better from a readability perspective. 

In this case for example, the first key is "id" which makes sense because that's the alert identifier and the second is "message" because that's the alert text. Then there are four time/date keys to specify the alert duration. Lastly there are optional meta-information keys that the client may not use like alert priority, routes, and signs. I'd personally rather have the data ordered like this instead of alphabetically which is completely arbitrary in terms of grouping and importance.

https://eslint.org/docs/rules/sort-keys
https://eslint.org/docs/rules/sort-vars

PR #79, opened 2018-10-03
fix routeNumber string issue in v1 API

zeladada:


young-k:
in the future, make sure to make this commit sentence case

Issue #56, opened 2018-09-16
Add multiroute endpoint


zeladada:
Endpoint to retrieve a route and ETA for multiple locations at a time.

There is no arriveBy parameter because the user should not ever need to specify a time to arrive by for multiple locations at a time.

----------
# **/multiroute** • GET

**Description**: Returns the best available route for each destination specified from a single start location

## Parameters

*required* **start** : String - “< latitude Double >,< longitude Double >”

| **description** | The starting point of the journey.                                                                                           |
| --------------- | ---------------------------------------------------------------------------------------------------------------------------- |
| **default**     | current location (set by client)                                                                                             |
| **notes**       | This can be a bus stop, a location found in Google Places (see Place ID docs), or coordinates (e.g. user’s current location) |

*required* **time** : Int

| **description** | The relevant time in the request.                                                                                                                                                                                                        |
| --------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **default**     | now (set by client)                                                                                                                                                                                                                      |
| **notes**       | If **arriveBy** is false, departBy functionality is used. The time is when the journey should at earliest begin by.
Otherwise, the time is when the route should arrive to the destination by

Time is in epoch (seconds since 1/1/1970) |

*required* **end** : [{Double, Double}] - “[{< latitude1 Double >,< longitude1 Double >}]”

| **description** | An array of latitude-longitude pair objects to specify ending pointsfa. Must contain at least one ending point. |
| --------------- | -------------------------------- |
| **default**     | n/a                              |
| **notes**       | See Start notes                  |

*required* **destinationNames** : [String] - “[< destinationName1 >]”

| **description** | The names of the destinations. Must contain at least one.                                                                    |
| --------------- | --------------------------------------------------------------------------------------------------------- |
| **default**     | n/a                                                                                                       |
| **notes**       | Each destination is used to change the final direction to the destination, as well as check if the destination is a bus stop. |

*optional*
**Add more destinations as needed, each must be in order and have an end location and name**

## Returns: [Route]

Returns an array of Routes, one for each destination.

*class* **Route**

| **Name**          | **Type**                                                                          | **Description**                                                                                 |
| ----------------- | --------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------- |
| departureTime     | String                                                                            | The time a user begins their journey (e.g. when to start walking)                               |
[...]

PR #93, opened 2018-10-27
Add multiroute endpoint

aliciaxw:
Query strings will have additional `end` and `destinationName` parameters that must be in order. #56 

ex. `&end=coords0&end=coords1&end=coords2&destinationName=name0&destinationName=name1&destinationName=name2`.

### DIFF
--- a/src/Api.js
+++ b/src/Api.js
@@ -29,9 +29,9 @@ class API extends AppDevAPI {
             AllStopsRouter,
             DelayRouter,
             HelloWorldRouter,
+            MultiRouteRouter,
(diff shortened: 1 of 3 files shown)


src/Api.js
```diff
@@ -30,6 +31,7 @@ class API extends AppDevAPI {
             HelloWorldRouter,
             PlacesAutocompleteRouter,
             RouteRouter,
+            MultiRouteRouter,
```
meganle: place this after `HelloWorldRouter` for alphabetical order

young-k: +1

PR #98, opened 2018-11-05
Replace register and improve logging

zeladada:
Initial PR for Analytics project.

Changes:

- replace register
- improved logging
- improved error handling
- chronicle integration and updates
- updated and cleaned-up packages
- various testing improvements
- finalized release-local diff compare function
- recurring data acquisition stability improvements and mutex
- simplified, improved webpack/build configs
- much faster dev builds
- more optimized prod builds
- simplified, improved jest/testing configs


src/Api.js
```diff
@@ -2,7 +2,7 @@
 import bodyParser from 'body-parser';
 import { Router } from 'express';
 
-import AppDevAPI from './appdev/AppDevAPI';
+import ApplicationAPI from './appdev/ApplicationAPI';
```
young-k: can we alphabetize this and move this down?

src/appdev/ChronicleSession.js
```diff
@@ -0,0 +1,77 @@
+// @flow
+import { config, S3 } from 'aws-sdk';
```
young-k: wait a second, why'd you copy & paste this in? i thought we were publishing `appdev` to `npm`?

src/routers/LoggingRouter.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+import type Request from 'express';
+import ApplicationRouter from '../appdev/ApplicationRouter';
+import LogUtils from '../utils/LogUtils';
```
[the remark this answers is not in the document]
young-k: address

src/utils/LogUtils.js
```diff
@@ -1,110 +1,104 @@
...
+const stop: ParquetSchema = { name: string, id: string };
+
+const direction: ParquetSchema = {
+    dirType: string,
+    name: string,
+    startTime: time,
+    endTime: time,
+    startLocation: pointObj,
+    endLocation: pointObj,
+    path,
```
young-k: this doesn't have an associated type?

src/test/TestUtils.js
```diff
@@ -116,25 +110,43 @@ async function printReleaseDiff(res, query) {
...
-                removeKeys(relJSON, ['path', 'stops']);
-                removeKeys(resCpy, ['path', 'stops']);
-                const delta = JsonDiff.diff(relJSON.data[0] || relJSON, resCpy);
+                releaseJSON = JSON.parse(releaseStr);
+                const responseJSONCopy = JSON.parse(responseStr);
+                const delta = JsonDiff.diff(releaseJSON, responseJSONCopy);
                 jsondiffpatch.console.log(delta);
             } catch (e) {
-                console.log(`WARNING: Current release: ${rel}\nLocal: ${JSON.stringify(res)}`);
+                console.log(`WARNING: Current release: ${releaseStr}\nLocal: ${JSON.stringify(res)}`);
```
meganle: You can use `responseStr` instead of `JSON.stringify(res)`

src/test/TestUtils.js
```diff
@@ -386,7 +398,8 @@ expect.extend(expectTests);
 
 export default {
     expectTests,
-    printDeep,
     getReleaseRes,
     printReleaseDiff,
+    logToFile,
```
meganle: alphabetize

src/test/TestUtils.js
zeladada: True. Nice catch

src/appdev/ChronicleSession.js
zeladada: More cryptic errors. I believe the issue is related to how Webpack is creating the bundle or how Jest runs tests on the bundle when used in conjunction with certain dependencies/libraries, same as the Axios bug we had (technically still have since we just replaced Axios altogether) that would crash the app because of an undefined variable error. I even wrote a custom branch of a library that `appdev` uses, added it, and wrote `appdev` tests to make sure it worked, but still no luck getting the `appdev` library to work with Transit. Then I more or less copy-pasted the files as you can see and it worked, so unfortunately I think it needs to stay like this unless someone is capable of fixing or even identifying the underlying issue that's causing certain dependencies to not work properly.

zeladada: To be clear the `appdev` project works fine it's just not working with Transit specifically.

PR #105, opened 2018-11-15
V2 API Refactoring

JunyoungLim:
All the major code changes are in the `ParseRouteUtils.js` file (i.e. `parseDetailedRoute` function).

V1 API returns the following:
* departureTime: Date
* arrivalTime: Date
* startCoords: Coordinates
* endCoords: Coordinates
* boundingBox: Bounds
* directions: [Direction]

V2 returns the following additional information:
* travelDistance: Double // distance between start and end coords in miles
* startName: String // starting location name
* endName: String // destination name
* totalDuration: Int // minutes between departureTime and arrivalTime
* routeSummary: [RouteSummaryElement]

Each `routeSummaryElement` consists of:
```javascript
{
    stopName: String,
    directionType: null || 'walk' || 'bus',
    routeNumber: null || Int,
    shouldStayOnBus: Bool,
}
```


src/ApiV2.js
```diff
@@ -0,0 +1,40 @@
...
+import MultiRouteRouter from './routers/MultiRouteRouter';
+import TrackingRouter from './routers/TrackingRouter';
+
+class APIv2 extends AppDevAPI {
+    getPath(): string {
+        return '/api/v2/';
+    }
+
+    middleware(): Array<any> {
+        return [
```
young-k: why not `return [bodyParser.json()];`?

src/routers/RouteRouterV2.js
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    // eslint-disable-next-line require-await
+    async content(req: Request): Promise<Array<Object>> {
+        const {
+            destinationName,
+            end,
+            start,
+            time: departureTimeQuery,
+            arriveBy,
```
young-k: you already know, alphabetize this

src/ApiV2.js
```diff
@@ -0,0 +1,40 @@
...
+import AlertsRouter from './routers/AlertsRouter';
+import AllStopsRouter from './routers/AllStopsRouter';
+import DelayRouter from './routers/DelayRouter';
+import HelloWorldRouter from './routers/HelloWorldRouter';
+import PlacesAutocompleteRouter from './routers/PlacesAutocompleteRouter';
+import RouteRouterV2 from './routers/RouteRouterV2';
+import MultiRouteRouter from './routers/MultiRouteRouter';
+import TrackingRouter from './routers/TrackingRouter';
+
+class APIv2 extends AppDevAPI {
```
young-k: this is inconsistent with the filename, you should stick with either `ApiV2` or `APIv2` but not have both!

src/utils/ParseRouteUtils.js
```diff
@@ -532,8 +534,137 @@ function parseRoute(resp: Object, destinationName: string) {
...
+ * @returns {Promise<Array>}
+ */
+async function parseDetailedRoute(resp: Object, destinationName: string) {
+    const routeInfo = await parseRoute(resp, destinationName);
+
+    const {
+        departureTime,
+        arrivalTime,
+        directions,
+        startCoords,
```
young-k: alphabetize

src/utils/ParseRouteUtils.js
```diff
@@ -532,8 +534,137 @@ function parseRoute(resp: Object, destinationName: string) {
...
+        { latitude: startCoords.lat, longitude: startCoords.long },
+        { latitude: endCoords.lat, longitude: endCoords.long }
+    );
+
+    travelDistance *= METERS_TO_MILES;
+    routeInfo.travelDistance = travelDistance;
+
+    let totalDuration = new Date(arrivalTime) - new Date(departureTime);
+    totalDuration /= 1000;
+    totalDuration /= 60;
```
young-k: can we make these constants as well?

src/utils/ParseRouteUtils.js
```diff
@@ -1,13 +1,15 @@
...
 import createGpx from 'gps-to-gpx';
+import geolib from 'geolib';
 import ErrorUtils from './LogUtils';
 import RequestUtils from './RequestUtils';
 import TCATUtils from './TCATUtils';
 import RealtimeFeedUtils from './RealtimeFeedUtils';
 import AllStopUtils from './AllStopUtils';
 
 const ONE_HOUR_MS = 3600000;
+const METERS_TO_MILES = 0.00062137119;
```
young-k: this should be alphabetized, so move this b4 the other one

src/utils/ParseRouteUtils.js
```diff
@@ -532,8 +534,137 @@ function parseRoute(resp: Object, destinationName: string) {
...
+            shouldStayOnBus: false,
+        });
+    } else {
+        let indexOfFirstBus = -1;
+        for (let i = 0; i < directions.length; i++) {
+            if (directions[i].type === 'depart') {
+                indexOfFirstBus = i;
+                break;
+            }
+        }
```
young-k: can `indexOfFirstBus` ever be `-1` after the `for` loop? if so, can we add some sort of error logging? if it's not possible, then why not just `let indexOfFirstBus;`?

src/utils/ParseRouteUtils.js
```diff
@@ -532,8 +534,137 @@ function parseRoute(resp: Object, destinationName: string) {
...
+
+        routeSummary.push({
+            stopName: directions[directions.length - 1].name,
+            direction: { type: null, routeNumber: null },
+            shouldStayOnBus: false,
+        });
+    }
+
+    routeInfo.routeSummary = routeSummary;
+
```
young-k: delete newline

src/utils/ParseRouteUtils.js
```diff
@@ -532,8 +534,137 @@ function parseRoute(resp: Object, destinationName: string) {
...
+    return routeInfo;
+}
+
 export default {
     parseWalkingRoute,
     parseRoute,
+    parseDetailedRoute,
```
young-k: alphabetize

src/utils/RouteUtils.js
```diff
@@ -91,7 +100,32 @@ async function getRoute(destinationName, end, start, departureTimeQuery, arriveB
...
+    // if there are no bus routes, we should just return walking instead of crashing
+    if (!busRoute && walkingRoute) {
+        return [walkingRoute];
+    }
+
+    // parse the graphhopper bus route
+    busRoute = await ParseRouteUtils.parseDetailedRoute(busRoute, destinationName);
+
+    // combine and filter to create the final route
+    busRoute = await createFinalRoute(
```
young-k: why not just `return await...` since you never use `busRoute` after it's assigned here?

src/utils/RouteUtils.js
```diff
@@ -91,7 +100,32 @@ async function getRoute(destinationName, end, start, departureTimeQuery, arriveB
...
+        departureTimeQuery,
+        arriveBy,
+    );
+
+    return busRoute;
+}
+
 export default {
     getRoute,
+    getDetailedRoute,
```
young-k: alphabetize :)

src/utils/ParseRouteUtils.js
```diff
@@ -532,8 +534,137 @@ function parseRoute(resp: Object, destinationName: string) {
...
+    /* routeSummaryDirection {
+        type: null || 'walk' || 'bus',
+        routeNumber: null || Int,
+    }
+    */
+
+    let routeSummary = [];
+
+    let walkOnlyRoute = true;
+    for (let i = 0; i < directions.length; i++) {
```
zeladada: You don't need to iterate through the directions here. In a walking route the route number and stops fields are null or empty which indicates walking only.

src/utils/ParseRouteUtils.js
```diff
@@ -532,8 +534,137 @@ function parseRoute(resp: Object, destinationName: string) {
...
+                    break;
+            }
+            routeSummary.push(newRouteSummary);
+        }
+
+        routeSummary.push({
+            stopName: directions[directions.length - 1].name,
+            direction: { type: null, routeNumber: null },
+            shouldStayOnBus: false,
+        });
```
zeladada: See condenseRoute in ParseRouteUtils, only one for-loop through the directions is done for all logic and parsing to simplify logic and save computation time. Therefore, place all this functionality within a single loop to avoid having to iterate through the directions more than once needlessly.

src/utils/RouteUtils.js
```diff
@@ -91,7 +100,32 @@ async function getRoute(destinationName, end, start, departureTimeQuery, arriveB
     return busRoute;
 }
 
+async function getDetailedRoute(destinationName, end, start, departureTimeQuery, arriveBy) {
```
zeladada: Please rename all "detailedRoute" to "routeV2" (for example: getDetailedRoute -> getRouteV2) or something similar so we know the object this function returns adheres to the V2 specification.

src/ApiV2.js
```diff
@@ -0,0 +1,40 @@
...
+        ];
+    }
+
+    routers(): Array<Router> {
+        return [
+            AlertsRouter,
+            AllStopsRouter,
+            DelayRouter,
+            HelloWorldRouter,
+            MultiRouteRouter,
```
zeladada: I just realized we're going to need a new multiroute router for V2. Should be somewhat easy, probably can just copy-paste MultiRouteRouter and switch out the V1 method calls for V2.

src/utils/ParseRouteUtils.js
```diff
@@ -532,8 +534,137 @@ function parseRoute(resp: Object, destinationName: string) {
...
+        directions,
+        startCoords,
+        endCoords,
+        boundingBox,
+        numberOfTransfers,
+    } = routeInfo;
+
+    let travelDistance = geolib.getDistance(
+        { latitude: startCoords.lat, longitude: startCoords.long },
+        { latitude: endCoords.lat, longitude: endCoords.long }
```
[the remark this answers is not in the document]
JunyoungLim: @young-k this is a function call instead of an object, do we still need to put comma?

src/utils/ParseRouteUtils.js
```diff
@@ -532,8 +534,137 @@ function parseRoute(resp: Object, destinationName: string) {
...
+            departureTime,
+            arrivalTime,
+            directions,
+            startCoords,
+            endCoords,
+        } = currPath;
+
+        let travelDistance = geolib.getDistance(
+            { latitude: startCoords.lat, longitude: startCoords.long },
+            { latitude: endCoords.lat, longitude: endCoords.long }
```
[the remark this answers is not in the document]
JunyoungLim: @young-k this is a function call instead of an object, do we still need to put a comma?

src/utils/ParseRouteUtils.js
```diff
@@ -532,8 +535,135 @@ function parseRoute(resp: Object, destinationName: string) {
...
+            let indexOfFirstBus = -1;
+            for (let i = 0; i < directions.length; i++) {
+                if (directions[i].type === 'depart') {
+                    indexOfFirstBus = i;
+                    break;
+                }
+            }
+
+            for (let i = indexOfFirstBus; i < directions.length; i++) {
+                const newRouteSummary = { stopName: directions[i].name, shouldStayOnBus: false };
```
myo3: Noticed that `shouldStayOnBus` is always set to `false`. Is there somewhere else in the code that you use the results from GraphHopper to check if you `shouldStayOnBus`? B/c from what I understand GraphHopper tells us that.

src/utils/ParseRouteUtils.js
```diff
@@ -532,8 +535,135 @@ function parseRoute(resp: Object, destinationName: string) {
...
+        currPath.totalDuration = Math.round(totalDuration);
+
+        currPath.startName = currPath.directions[0].name;
+        currPath.endName = destinationName;
+
+        // routeSummary
+
+        /* routeSummaryElement {
+            stopName: String,
+            direction: RouteSummaryDirection?,
```
myo3: Is`RouteSummaryDirection?` supposed to have a question mark? B/c in the PR description you write `RouteSummaryDirection` which indicates every `routeSummaryElement` should have a `RouteSummaryDirection`

src/utils/ParseRouteUtils.js
```diff
@@ -532,8 +535,135 @@ function parseRoute(resp: Object, destinationName: string) {
...
+                const newRouteSummary = { stopName: directions[i].name, shouldStayOnBus: false };
+                switch (directions[i].type) {
+                    case 'depart':
+                        newRouteSummary.direction = { type: 'bus', routeNumber: directions[i].routeNumber };
+                        break;
+                    case 'walk':
+                        newRouteSummary.direction = { type: 'walk', routeNumber: null };
+                        break;
+                    default:
+                        break;
```
myo3: should you be doing `newRouteSummary.direction = { type: null, routeNumber: null }` for the case where there's no `direction`? b/c i notice you fill in the `direction` field of a routeSummaryElement with the default `{ type: null, routeNumber: null }` object for last direction in the `if (walkOnlyRoute)` case.

Truthfully tho this would be an error if any routeSummary object, except the last one, had no `direction`. So maybe you should log that

src/utils/ParseRouteUtils.js
```diff
@@ -532,8 +535,135 @@ function parseRoute(resp: Object, destinationName: string) {
...
+                }
+                routeSummary.push(newRouteSummary);
+            }
+
+            routeSummary.push({
```
myo3: aren't u pushing the last direction twice (`directions[directions.length - 1]`)? b/c you iterate through the directions to push on `newRouteSummary` to `routeSummary` and that for-loop goes through the last direction: `for (let i = indexOfFirstBus; i < directions.length; i++)`, but then you're also do this `push` where you push `directions[directions.length - 1].name`. Unless this is supposed to be the case?

young-k:
Can you also squash the last few commits into the last one?

`git rebase -i HEAD~4` and `squash` all of them except the oldest (it'll squash them into the oldest), and then push to this branch with `git push --force`

src/utils/ParseRouteUtils.js
```diff
@@ -532,8 +534,130 @@ function parseRoute(resp: Object, destinationName: string) {
...
+
+        /* routeSummaryDirection {
+            type: null || 'walk' || 'bus',
+            routeNumber: null || Int,
+        }
+        */
+
+        const routeSummary = [];
+
+        const walkOnlyRoute = directions.length === 0 || !directions[0].routeNumber;
```
JunyoungLim: @zeladada will this be a sufficient check for `walkOnlyRoute`? I removed the for-loop.

src/routers/LoggingRouter.js
```diff
@@ -0,0 +1,21 @@
...
+class LoggingRouter extends ApplicationRouter<string> {
+    constructor() {
+        super('GET');
+    }
+
+    getPath(): string {
+        return '/logSelection/';
+    }
+
+    async content(req: Request): Promise<string> {
```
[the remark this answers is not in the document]
young-k: address both comments, they're valid @JunyoungLim

src/appdev/ChronicleSession.js
```diff
@@ -0,0 +1,77 @@
...
+const PATH = '/tmp';
+const BUCKET = 'appdev-register';
+
+class ChronicleSession {
+    app: string;
+
+    cacheSize: number;
+
+    logMap: Map<string, Object[]>;
+
```
young-k: delete newlines in between every var

src/utils/LogUtils.js
```diff
@@ -1,110 +1,104 @@
...
+    CHRONICLE_APP_NAME,
+);
+
+// define common parquet types
+const time = { type: 'TIMESTAMP_MILLIS' }; // INT64 - epoch time
+const boolean = { type: 'BOOLEAN' }; // BOOLEAN
+const string = { type: 'UTF8' }; // BYTE_ARRAY
+const float = { type: 'FLOAT' }; // FLOAT32 - is within 1.7m lat/long precision
+const JSON = { type: 'JSON' }; // BYTE_ARRAY - encoded JSON obj
+const int = { type: 'INT32' }; // INT32
```
young-k: alphabetize

src/utils/LogUtils.js
```diff
@@ -130,100 +124,75 @@ function log(obj: Object, error: ?boolean = false) {
...
     log,
     logErr,
-    logToFile,
-    printDeep,
+    writeToFile,
+    logToChronicle,
+    routeRequestSchema,
+    routeResultSchema,
+    userSelectionSchema,
+    cacheSchema,
```
young-k: alphabetize

PR #120, opened 2018-12-06
Refactor RequestUtils.js

young-k:
Verified nothing broke:

image from ios 2

Verifying that I replaced all instances of renamed functions:

screen shot 2018-12-06 at 6 56 23 pm

### DIFF
--- a/src/utils/RequestUtils.js
+++ b/src/utils/RequestUtils.js
@@ -6,7 +6,7 @@ import util from 'util';
 import LogUtils from './LogUtils';
 
 function createRequest(
-    options: any,
+    options: Object,
     errorMessage: ?string = 'Request failed',
     verbose: ?boolean = false,
     returnRes: ?boolean = false,
@@ -29,9 +29,17 @@ function createRequest(
     });
 }
 
-async function fetchRetry(fn: () => void, n: number = 5): any {
+/**
+ * Attempts to call [fn] [retryCount] times. Only if an exception occurs [retryCount] times, throw the exception.
+ * Else returns the result of [fn].
+ *
+ * @param {() => any} fn
+ * @param {number} retryCount
+ * @returns {any} The result of [fn]
+ */
+async function fetchWithRetry(fn: () => any, retryCount: number = 5): any {
     let error;
-    for (let i = 0; i < n; i++) {
+    for (let i = 0; i < retryCount; i++) {
         try {
             // eslint-disable-next-line no-await-in-loop
             return (await fn());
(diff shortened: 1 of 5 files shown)


src/utils/RequestUtils.js
```diff
@@ -29,9 +29,17 @@ function createRequest(
...
-async function fetchRetry(fn: () => void, n: number = 5): any {
+/**
+ * Attempts to call [fn] [retryCount] times. Only if an exception occurs [retryCount] times, throw the exception.
+ * Else returns the result of [fn].
+ *
+ * @param {() => any} fn
+ * @param {number} retryCount
+ * @returns {any} The result of [fn]
+ */
+async function retryFetch(fn: () => any, retryCount: number = 5): any {
```
kq-li: prefer `fetchWithRetry`

src/utils/RequestUtils.js
```diff
@@ -43,28 +51,39 @@ async function fetchRetry(fn: () => void, n: number = 5): any {
...
+ * @param {number} refreshInterval Ms to wait before refresh
+ * @param {number} timeout Ms to wait before request times out
+ * @param {Object} objectToUpdate
  */
-function startRequestIntervals(fn: () => void, refreshInterval: number, timeout: number, setObj: Object): void {
+function updateObjectOnInterval(
+    fn: () => void,
+    refreshInterval: number,
+    timeout: number,
+    objectToUpdate: Object,
```
kq-li: should be `any` since `fn` returns any

src/utils/RequestUtils.js
```diff
@@ -43,28 +51,39 @@ async function fetchRetry(fn: () => void, n: number = 5): any {
...
+ * If [fn] completes successfully, [objectToUpdate] is set to the value of [fn()].
+ *
+ * @param {() => any} fn
+ * @param {number} refreshInterval Ms to wait before refresh
+ * @param {number} timeout Ms to wait before request times out
+ * @param {Object} objectToUpdate
  */
-function startRequestIntervals(fn: () => void, refreshInterval: number, timeout: number, setObj: Object): void {
+function updateObjectOnInterval(
+    fn: () => void,
```
kq-li: should be `() => any`

src/utils/RequestUtils.js
young-k: aiya ok

src/utils/RequestUtils.js
young-k: should be `Object` right?

PR #135, opened 2019-01-29
Add live tracking microservice

young-k:


src/routers/DelayRouter.js
```diff
@@ -14,13 +14,13 @@ class DelayRouter extends ApplicationRouter<Object> {
 
     async content(req): Promise<any> {
         const { stopID, tripID } = req.query;
-        LogUtils.log({ category: 'DelayRouter', stopID, tripID });
-        const res = RealtimeFeedUtils.getTrackingInformation(
+        LogUtils.log({
+            category: 'DelayRouterRTF',
```
meganle: probably should just change to 'DelayRouter' since we're not returning the rtf
```suggestion
            category: 'DelayRouter',
```

Issue #102, opened 2018-11-11
Route sorting based on inputs


mattbarker016:
Design proposed a new way to sort bus routes. Below is a in-progress screenshot. We want to do this sorting on the backend to avoid duplicate logic.

<img width="122" alt="image" src="https://user-images.githubusercontent.com/8039928/48316510-99148b80-e5b2-11e8-81ae-6ce3d36cd1be.png">

### General

**Section 1: Boarding Soon**
All other bus routes that involve different "from" bus stops. Sorted to get people there the fastest.

**Section 2: Walking Routes**

### Edge Case: Leaving from Bus Stop
_This is only when the "from" location is a specific bus stop._

**Section 1: From [Bus Stop Name]**
Routes that _only_ leave from that specific bus stop.

**Section 2: Boarding Soon**
All other bus routes with different "from" bus stops. Sorted by fastest arrival time.

**Section 3: Walking Routes**

## Response

Data will map to a JSON object instead of an array of routes. The key of each field should be the name of the section, which client-side will read directly to access the name of the stop. The value will be an array of routes. Work with Maya on design for the proper language.

```
{
    "data" : {
       "From Ithaca Commons - Seneca St." : [
             { "route1" : ... } , ...
        ],
        "Boarding Soon" : [
             { "route2" : ... } , ...
        ],
        "Walking Routes" : [
             { "route3" : ... } , ...
        ],
    }
}
```

young-k:
@kevinchan159 @alanna-zhou the next step is to implement the **Edge Case: Leaving from Bus Stop** section, I've already added functionality so that walking routes are always returned no matter what!

kevinchan159:
@mattbarker016 we're going to change the format of the response to something like this since you probably don't want dynamic keys
```
{
    "data" : {
       "fromStop" : [
             { "route1" : ... } , ...
        ],
        "boardingSoon" : [
             { "route2" : ... } , ...
        ],
        "walking" : [
             { "route3" : ... } , ...
        ],
    }
}
```

mattbarker016:
Yeah, that makes sense actually, my bad!

PR #146, opened 2019-01-31
Always add walking routes at the bottom of results

young-k:
More fixes:

- Some progress for #102, walking routes are always appended
- Fixed bug where walking route would have a start time of 5 minutes earlier than the search
- Decreased delay for startup of server if `RTF` is not running
- Refactored code in `RouteUtils`

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -13,7 +13,7 @@ LogUtils.log({ message: 'server.js: Initializing data and waiting for Graphhoppe
 
 const PORT: number = parseInt(process.env.PORT) || 80;
 const SERVER_ADDRESS: string = '0.0.0.0';
-const TWENTY_SECONDS_IN_MS: number = 20000;
+const FIVE_SECONDS_IN_MS: number = 5000;
 
 const authToken = TokenUtils.fetchAuthHeader();
 
@@ -25,7 +25,7 @@ const { express } = app;
 const init = new Promise((resolve, reject) => {
     // start endpoints that rely on external data starting with authentication token
     const timeoutPromise = new Promise((res, rej) => {
-        setTimeout(res, TWENTY_SECONDS_IN_MS);
+        setTimeout(res, FIVE_SECONDS_IN_MS);
     }).then(value => LogUtils.log({ message: 'server.js: Timeout reached' }));
     authToken.then(() => {
         // await data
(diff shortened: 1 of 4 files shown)


src/server.js
```diff
@@ -13,7 +13,7 @@ LogUtils.log({ message: 'server.js: Initializing data and waiting for Graphhoppe
 
 const PORT: number = parseInt(process.env.PORT) || 80;
 const SERVER_ADDRESS: string = '0.0.0.0';
-const TWENTY_SECONDS_IN_MS: number = 20000;
+const FIVE_SECONDS_IN_MS: number = 5000;
```
alanna-zhou: why do you have to save this in a constant? if you were going to abstract out a timeout time why wouldn't you name it something other than the word description of the numerical value?

src/utils/RouteUtils.js
```diff
@@ -15,8 +14,14 @@ import ParseRouteUtils from './ParseRouteUtils';
  * @param arriveBy
  * @returns {*}
  */
-/* eslint-disable no-param-reassign */
-async function createFinalRoute(routeBus, routeWalking, start, end, departureTimeQuery, arriveBy: boolean) {
+async function createFinalRoute(
+    routeBus: Array<Object>,
+    routeWalking: Object,
```
alanna-zhou: why not just routeWalk since you do routeBus instead of routeBusing

src/utils/RouteUtils.js
```diff
@@ -15,8 +14,14 @@ import ParseRouteUtils from './ParseRouteUtils';
...
+    start: string,
+    end: string,
+    departureTimeQuery: number,
+    arriveBy: boolean,
```
alanna-zhou: what is arriveBy and why is it a boolean? is it like if the user arrived or not? cuz arriveby sounds like it should be a time

src/server.js
young-k: it's to explain what the value corresponds to, because if someone saw `5000` they might not be sure what 5000 represents -> this makes it explicit that it's referring to a specific number of milliseconds!

it's the difference between `f(5000)` and `f(FIVE_SECONDS_IN_MS)`, the second one is much more readable

src/utils/RouteUtils.js
young-k: difficult to change this because it's in production with this name already on the client-side, changing it will make it more difficult to debug errors

Issue #145, opened 2019-01-31
Deprecate fetch of stops & alerts


young-k:
We should deprecate these fetches in the node server once it's migrated to the new microservice.

PR #147, opened 2019-02-01
Deprecate fetch of stops & alerts #145

kevinchan159:
Issue #145 
- Replaced old code of fetching stops & alerts with new code to fetch both from the `transit-microservice`

`NOTE`: Not sure if somethings wrong w/ my linter? It seems to have added/removed some spaces in certain lines. Is my linter correct?

Screenshots of responses for `alerts` & `allStops`:
Alerts:
screen shot 2019-01-31 at 10 29 21 pm

All Stops:
screen shot 2019-01-31 at 10 29 13 pm

### DIFF
--- a/env.template
+++ b/env.template
@@ -1,11 +1,9 @@
-ALERTS=localhost
-ALL_STOPS=localhost
 GHOPPER_BUS=localhost
 GHOPPER_WALKING=localhost
-LIVE_TRACKING=localhost
 MAP_MATCHING=localhost
 NODE_ENV=production
 PLACES_KEY=PLACES_KEY
 PORT=3000
 RELEASE_URL=http://localhost:3001/
+TCAT_SERVICE=localhost
 TOKEN=token
--- a/env.template
+++ b/env.template
@@ -4,6 +4,6 @@ MAP_MATCHING=localhost
 NODE_ENV=production
 PLACES_KEY=PLACES_KEY
 PORT=3000
+PYTHON_APP=localhost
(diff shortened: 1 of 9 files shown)


env.template
```diff
@@ -1,3 +1,5 @@
+ALERTS=localhost
+ALL_STOPS=localhost
```
young-k: we should just call this `TCAT_SERVICE` instead of having one per alerts, all_stops, and live_tracking

src/utils/AlertsUtils.js
```diff
@@ -1,94 +1,17 @@
...
-            return 'Saturday';
-        case 1:
-            return 'Sunday';
-        default:
-            return '';
-    }
+async function fetchAlerts(): Object {
+    const options = {
+        method: 'GET',
+        url: `http://${ALERTS || 'localhost'}:5000/alerts`,
```
young-k: update this and the other ones

env.template
```diff
@@ -1,9 +1,9 @@
 GHOPPER_BUS=localhost
 GHOPPER_WALKING=localhost
-LIVE_TRACKING=localhost
 MAP_MATCHING=localhost
 NODE_ENV=production
 PLACES_KEY=PLACES_KEY
 PORT=3000
 RELEASE_URL=http://localhost:3001/
+TCAT_SERVICE=localhost
```
young-k: actually let's call this `PYTHON_APP`, to make sure it's super clear

PR #152, opened 2019-02-04
Fix RouteRouter & MultiRouteRouter

kevinchan159:
Issue #151, #153 
- Changed `ApplicationRouter` to support multiple request types
- Fix up `MultiRouteRouter` and updated README with updated spec.
- Changed `RouteRouter` to support `POST`

cc @ywsang

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,6 +243,7 @@ Otherwise, the time is when the route should arrive to the destination by
 Time is in epoch (seconds since 1/1/1970) |
 
 *required* **end** : [String] - “[“< latitude Double >,< longitude Double >”]”
+
 | **description** | An array of latitude-longitude strings to specify ending points. Must contain at least one ending point. Should be the same length as **destinationNames**. |
 | --------------- | -------------------------------- |
 | **default**     | n/a                              |
--- a/README.md
+++ b/README.md
@@ -242,9 +242,8 @@ Otherwise, the time is when the route should arrive to the destination by
 
 Time is in epoch (seconds since 1/1/1970) |
 
-*required* **end** : [{Double, Double}] - “[{< latitude1 Double >,< longitude1 Double >}]”
-
-| **description** | An array of latitude-longitude pair objects to specify ending points. Must contain at least one ending point. |
+*required* **end** : [String] - “[“< latitude Double >,< longitude Double >”]”
+| **description** | An array of latitude-longitude strings to specify ending points. Must contain at least one ending point. Should be the same length as **destinationNames**. |
 | --------------- | -------------------------------- |
 | **default**     | n/a                              |
 | **notes**       | See Start notes                  |
--- a/src/appdev/ApplicationRouter.js
+++ b/src/appdev/ApplicationRouter.js
@@ -36,14 +36,14 @@ class AppDevResponse<T> {
 class ApplicationRouter<T> {
     router: Router;
 
-    requestType: RequestType;
+    requestTypes: Array<RequestType>;
 
     /**
      * Subclasses must call this constructor and pass in the HTTP method
      */
-    constructor(type: RequestType) {
+    constructor(types: Array<RequestType>) {
         this.router = new Router();
-        this.requestType = type;
+        this.requestTypes = types;
 
         // Initialize this router
         this.init();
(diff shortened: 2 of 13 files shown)


src/utils/RouteUtils.js
```diff
@@ -86,7 +85,7 @@ async function getRoutes(
         start,
         end,
         departureTimeQuery,
-        arriveByBool,
+        arriveBy,
```
alanna-zhou: if arriveByBool is getting changed to arriveBy anyway could it be changed to isArriveBy

README.md
```diff
@@ -242,9 +242,9 @@ Otherwise, the time is when the route should arrive to the destination by
 
 Time is in epoch (seconds since 1/1/1970) |
 
-*required* **end** : [{Double, Double}] - “[{< latitude1 Double >,< longitude1 Double >}]”
+*required* **end** : [String] - “[“< latitude Double >,< longitude Double >”]”
```
young-k: can we check with iOS, because this should optimally be an array of float tuples –– I really want to convert this to a POST request, so that we get better encoding of booleans and floats, and we don't have to do this manually on the server cc @ywsang

src/routers/MultiRouteRouter.js
```diff
@@ -22,21 +22,21 @@ class MultiRouteRouter extends ApplicationRouter<Array<Object>> {
...
+        // each destinationName should correspond to one end point
+        if (destinationNames.length !== end.length) {
+            return [];
         }
 
         // multiple destinations given
         const routes = [];
-        for (let i = 0; i < destinationName.length; i++) {
-            routes.push(RouteUtils.getRoutes(destinationName[i], end[i], start[i], departureTimeQuery, false));
+        for (let i = 0; i < destinationNames.length; i++) {
```
young-k: why not use a `map` operation? it's cleaner

src/routers/RouteRouter.js
```diff
@@ -26,7 +26,8 @@ class RouteRouter extends ApplicationRouter<Array<Object>> {
             uid,
         } = req.query;
 
-        const routes = await RouteUtils.getRoutes(destinationName, end, start, departureTimeQuery, arriveBy);
+        const arriveByBool = arriveBy === '1';
+        const routes = await RouteUtils.getRoutes(destinationName, end, start, departureTimeQuery, arriveByBool);
```
young-k: I think we should make a version of `RouteRouter` so that this is a POST request as well, same comment regarding better boolean encoding. We can create it on the backend first, and client can push it out in a following update –– can we make a Git issue for this?

src/utils/RouteUtils.js
kevinchan159: yea good point

src/routers/MultiRouteRouter.js
kevinchan159: yup will do

src/appdev/ApplicationRouter.js
```diff
@@ -60,19 +60,21 @@ class ApplicationRouter<T> {
...
+                    this.router.get(path, this.response());
+                    break;
+                case 'POST':
+                    this.router.post(path, this.response());
+                    break;
+                case 'DELETE':
+                    this.router.delete(path, this.response());
+                    break;
+                default:
+                    throw new Error('HTTP method not specified!');
```
young-k: shouldn't this be `HTTP method not supported`?

src/routers/MultiRouteRouter.js
```diff
@@ -21,23 +21,27 @@ class MultiRouteRouter extends ApplicationRouter<Array<Object>> {
     // Request does not require an arriveBy query param, unlike in RouteRouter
     // eslint-disable-next-line require-await
     async content(req: Request): Promise<Array<Object>> {
+        const params = req.method === 'GET' ? req.query : req.body;
```
young-k: wow congratulations this is clean stuff

src/routers/RouteRouter.js
```diff
@@ -24,9 +25,10 @@ class RouteRouter extends ApplicationRouter<Array<Object>> {
             start,
             time: departureTimeQuery,
             uid,
-        } = req.query;
+        } = params;
 
-        const routes = await RouteUtils.getRoutes(destinationName, end, start, departureTimeQuery, arriveBy);
+        const arriveByBool = (arriveBy === '1' || arriveBy === true);
```
young-k: `isArriveBy`?

Issue #149, opened 2019-02-02
Integrate ceiling of at least 1 min for destination times


alanna-zhou:
So that routes don't show start and end times as the same

PR #159, opened 2019-02-11
Add 1 minute time difference check

kevinchan159:
Issue #149 

- Ensure that routes always have a minimum one minute difference

Verified on dev server

Screenshots:
52444227_2467266743501327_2349293887677267968_n
51666596_347948599137613_590690661854871552_n

### DIFF
--- a/src/utils/ParseRouteUtils.js
+++ b/src/utils/ParseRouteUtils.js
@@ -9,6 +9,7 @@ import RealtimeFeedUtils from './RealtimeFeedUtils';
 import RequestUtils from './RequestUtils';
 
 const ONE_HOUR_IN_MILLISECONDS = 3600000;
+const ONE_MIN_IN_MILLISECONDS = 60000;
 
 /**
  * distanceBetweenPoints(point1, point2) returns the distance between two points in miles
@@ -180,7 +181,7 @@ async function condenseRoute(
                  * far away with no walking direction in between, EG: 1. get off at Statler 2. board at RPCC
                  */
                 if (previousDirection.stops[previousDirection.stops.length - 1].stopID
-                !== direction.stops[0].stopID) {
+                    !== direction.stops[0].stopID) {
                     return null;
                 }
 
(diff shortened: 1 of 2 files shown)


src/utils/ParseRouteUtils.js
```diff
@@ -222,9 +223,36 @@ async function condenseRoute(
...
+        ).toISOString().split('.')[0]}Z`;
+    }
+}
+
+/**
+ * Returns whether departureDate and arrivalDate are within the same minute
+ * @param route
+ * @returns boolean
+ */
+function isValidDates(departureDate: Date, arrivalDate: Date): boolean {
```
young-k: this function isn't named properly, `isValidDates`? it should describe what the function is doing
also `@param route` isn't correct either

src/utils/ParseRouteUtils.js
```diff
@@ -222,9 +223,36 @@ async function condenseRoute(
...
+    return !(departureDate.getTime() + ONE_MIN_IN_MILLISECONDS > arrivalDate.getTime()
```
young-k: why is the first check necessary?

kevinchan159: the minutes could be the same but the hour is different

young-k: ah you are smart, can we make these explicit as in:
```
const isSameMinute = ...
const isSameHour = ...
return isSameMinute && isSameHour;
```
it'll be easier to understand in the future

young-k:
In the future, you should run `git commit --amend`, and `git push --force`, rather than making a new commit to address the feedback!

src/utils/ParseRouteUtils.js
```diff
@@ -222,9 +223,38 @@ async function condenseRoute(
...
+    }
+}
+
+/**
+ * Returns whether departureDate and arrivalDate are within the same minute
+ * @param departureDate
+ * @param arrivalDate
+ * @returns boolean
+ */
+function datesAreSameMinute(departureDate: Date, arrivalDate: Date): boolean {
```
young-k: maybe `areDatesTheSameMinute`? the more verbose the better

kevinchan159:
> In the future, you should run `git commit --amend`, and `git push --force`, rather than making a new commit to address the feedback!

I was gonna squash all of them after @alanna-zhou reviews it in case I need to change anything else

PR #161, opened 2019-02-12
Re-dockerize application; remove wait-on

young-k:
I've simplified the development flow significantly, as noted in the new README:

```
npm run start:microservices # will start all other microservices
npm run start:node # will start the node server
```

We have *hot-reloading* through Docker, thank me later:

What does this mean? We can make changes locally, these will be persisted into the Docker container, and the Docker container will automatically update and re-compile with the new changes.

screen shot 2019-02-11 at 11 11 12 pm

Proof that it's working locally:
screen shot 2019-02-11 at 11 10 26 pm


src/utils/GraphhopperUtils.js
```diff
@@ -138,9 +109,7 @@ function getDepartureTimeDateNow(departureTimeQuery: string, isArriveByQuery: bo
  * @param isArriveByQuery
  * @returns {Promise<{busRoute: any, walkingRoute: any}>}
  */
-async function fetchRoutes(end: string, start: string, departureTimeDateNow: string, isArriveByQuery: boolean) {
-    const isReady = await isGraphhopperReady();
-    if (!isReady) return null;
+async function fetchRoutes(end: string, start: string, departureTimeDateNow: string, isArriveByQuery: boolean): Object {
```
alanna-zhou: why is the variable `departureTimeDateNow`like is there some `departureTimeDateLater` or sth

src/server.js
```diff
@@ -25,20 +24,14 @@ const init = new Promise((resolve, reject) => {
     const timeoutPromise = new Promise((res, rej) => {
         setTimeout(res, FIVE_SECONDS_IN_MS);
     }).then(value => LogUtils.log({ message: 'server.js: Timeout reached' }));
+
     authToken.then(() => {
         // await data
-        const dataInit = Promise.race([
+        Promise.race([
```
alanna-zhou: why don't you need to save this `dataInit` anymore

Dockerfile
```diff
@@ -1,15 +1,19 @@
...
+COPY . .
+ENV NODE_ENV development
+EXPOSE 3000
+CMD npm run start:dev
 
+# PRODUCTION
+FROM node:10.15-alpine AS release
+WORKDIR /usr/src/app
 COPY . .
+COPY --from=development /usr/src/app/node_modules /usr/src/app/node_modules
```
kq-li: what even

src/server.js
meganle: before it was used on line 39, but that isn't the case anymore

young-k: ^^ basically I deprecated `GhopperUtils.isGraphhopperReady`, and then there's only one Promise in the `Promise.all`, so that can be removed completely and the `then` clause can be moved into the original `then` clause of `dataInit`

src/utils/GraphhopperUtils.js
young-k: will investigate this

PR #166, opened 2019-02-14
Combine bus stop and google place searches into one endpoint

alanna-zhou:
Verified on dev server

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,7 +249,41 @@ Returns an array of Routes, one for each destination.
 | boundingBox       | {  maxLat: Double,  maxLong: Double,  minLat: Double,  minLong: Double  } | The most extreme points of the route. Used to center the map on client-side (with some padding) |
 | numberOfTransfers | Int                                                                               | Number of transfers in route. Default 0.    
 
+----------
+# **/search** • POST
 
+**Description**: Returns a list of bus stops and google autocomplete search results given a query string.
+
+## Parameters
+
+*required* **query** : String
+
+| **description** | The user's search query for a bus stop or place.                                                                                  |
+| --------------- | ---------------------------------------------------------------------------------------------------------------------------- |
+| **notes**       | This can be any string the user types in. |
+
+## Returns: [JSON]
+
+Returns an array of places, which can identify as either "busStop" or "googlePlace".
+
+**busStop**
+
+| **Name** | **Type** | **Description**                           |
+| -------- | -------- | ----------------------------------------- |
+| type     | "busStop" | The type of search suggestion. |
+| name     | String   | The name of the bus stop.                 |
+| lat      | Double   | The latitude coordinate of the bus stop.  |
+| long     | Double   | The longitude coordinate of the bus stop. |
+
+
+**googlePlace**
+
+| **Name** | **Type** | **Description**                           |
+| -------- | -------- | ----------------------------------------- |
+| type     | "googlePlace" | The type of search suggestion. |
+| detail     | String   | The address of the place.                |
+| name      | Double   | The name of the place.  |
+| placeID    | Double   | The Google place ID. |
 
 ----------
 # **/tracking** • POST
--- a/src/Api.js
+++ b/src/Api.js
@@ -2,16 +2,7 @@
 import bodyParser from 'body-parser';
 
 import ApplicationAPI from './appdev/ApplicationAPI';
-import AlertsRouter from './routers/AlertsRouter';
-import AllStopsRouter from './routers/AllStopsRouter';
-import DelayRouter from './routers/DelayRouter';
-import HelloWorldRouter from './routers/HelloWorldRouter';
-import PlacesAutocompleteRouter from './routers/PlacesAutocompleteRouter';
-import RouteRouter from './routers/RouteRouter';
-import SearchRouter from './routers/SearchRouter';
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -249,7 +249,41 @@ Returns an array of Routes, one for each destination.
...
 | numberOfTransfers | Int                                                                               | Number of transfers in route. Default 0.    
 
+----------
+# **/search** • POST
+
+**Description**: Returns a list of bus stops and google autocomplete search results given a query string.
+
+## Parameters
+
+*required* **query** : String - “st”
```
meganle: I don't think you need the "st" (Look at the docs for like `/tracking`)/ The param is straight forward unlike lat, lng coords so it's fine to just put String

src/routers/SearchRouter.js
```diff
@@ -0,0 +1,79 @@
+// @flow
```
meganle: Looks like this file already existed and was just renamed. Neat trick for next time: when you rename a file, use `git mv` that way git will recognize that the file was renamed and just show the changes that were actually made to the file!

src/Api.js
```diff
@@ -8,6 +8,7 @@ import DelayRouter from './routers/DelayRouter';
 import HelloWorldRouter from './routers/HelloWorldRouter';
 import PlacesAutocompleteRouter from './routers/PlacesAutocompleteRouter';
 import RouteRouter from './routers/RouteRouter';
+import SearchRouter from './routers/SearchRouter';
 import MultiRouteRouter from './routers/MultiRouteRouter';
```
young-k: can we make an `index.js` in routers so we can just import all of these from `./routers/`, instead of each of them individually?

src/routers/SearchRouter.js
young-k: genius

src/routers/SearchRouter.js
```diff
@@ -0,0 +1,79 @@
...
+import LRU from 'lru-cache';
+import AllStopUtils from '../utils/AllStopUtils';
+import ApplicationRouter from '../appdev/ApplicationRouter';
+import RequestUtils from '../utils/RequestUtils';
+
+const BUS_STOP = 'busStop';
+const cacheOptions = {
+    max: 10000,
+    maxAge: 1000 * 60 * 60 * 24 * 5,
```
young-k: can we add comments for what these represent

src/routers/SearchRouter.js
```diff
@@ -0,0 +1,79 @@
...
+            const googlePredictions = predictions.map(p => ({
+                type: GOOGLE_PLACE,
+                detail: p.structured_formatting.secondary_text,
+                name: p.structured_formatting.main_text,
+                placeID: p.place_id,
+            }));
+            cache.set(query, googlePredictions);
+
+            const places = formattedStops.concat(googlePredictions);
+            return places;
```
young-k: we shouldn't define a `const` to return it on the next line

src/utils/RouteUtils.js
```diff
@@ -70,6 +70,9 @@ async function getRoutes(
         destinationName,
     );
 
+    // Ensure that arrivalTime and departureTime have at least one minute difference
+    ParseRouteUtils.adjustRouteTimesIfNeeded(walkingRoute);
```
young-k: isn't this from @kevinchan159  changes? can we merge `master` into this branch so this doesn't show up in your changes?

```
git checkout master
git pull
git checkout [YOUR BRANCH]
git merge master # merges master into the current branch
```

src/routers/SearchRouter.js
```diff
@@ -0,0 +1,79 @@
...
+
+        // not in cache
+        const options = {
+            method: 'GET',
+            url: 'https://maps.googleapis.com/maps/api/place/autocomplete/json',
+            qs: {
+                input: query,
+                key: process.env.PLACES_KEY,
+                location: '42.4440,-76.5019',
+                radius: 24140,
```
young-k: why is this the radius?

src/routers/SearchRouter.js
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        // not in cache
+        const options = {
+            method: 'GET',
+            url: 'https://maps.googleapis.com/maps/api/place/autocomplete/json',
+            qs: {
+                input: query,
+                key: process.env.PLACES_KEY,
+                location: '42.4440,-76.5019',
```
young-k: can we define this in a const up top

src/routers/SearchRouter.js
```diff
@@ -0,0 +1,79 @@
...
+class SearchRouter extends ApplicationRouter<string> {
+    constructor() {
+        super(['POST']);
+    }
+
+    getPath(): string {
+        return '/search/';
+    }
+
+    async content(req): Promise<any> {
```
young-k: can we make the type stronger than `Promise<any>`, and add a type for `req` as well?

src/routers/SearchRouter.js
kevinchan159: I made it a const so that people know what the array refers too. I'll just add a comment instead then

src/routers/SearchRouter.js
kevinchan159: This was taken from `PlacesAutocompleteRouter`

PR #167, opened 2019-02-17
Change to 2 spaces

kevinchan159:
Change to 2 spaces


kevinchan159:
> what's with the hound issues

its a rule that we don't wont to have. i turned it off in the latest commit

PR #170, opened 2019-02-18
Remove tcat-ny-us/ from .gitignore

young-k:


young-k:
I thought we don't need this at all anymore, with the changes pushed by you and @alanna-zhou?

Issue #168, opened 2019-02-18
Add fuzzy string matching to `/search` Endpoint


kevinchan159:

kevinchan159:
Resolved in #172

Issue #169, opened 2019-02-18
Rearrange ordering of places in `/search` endpoint


kevinchan159:
Order should be:
- Max of 5 google places
- Bus stops that pass the Levenstein filter

kevinchan159:
Resolved in #172

PR #172, opened 2019-02-20
Integrate fuzzy search for bus stops and return google places first

alanna-zhou:


src/routers/SearchRouter.js
```diff
@@ -69,10 +79,11 @@ class SearchRouter extends ApplicationRouter<Array<Object>> {
         name: p.structured_formatting.main_text,
         placeID: p.place_id,
       }));
-      cache.set(query, googlePredictions);
+      const filteredPredictions = googlePredictions.filter(p => formattedStops.find(s => p.name.includes(s.name)) === undefined);
```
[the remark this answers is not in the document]
young-k: This line is extremely difficult to understand, defining a function above just to pass into here with some comments will save time in the future for developers!

src/routers/SearchRouter.js
```diff
@@ -32,7 +34,15 @@ class SearchRouter extends ApplicationRouter<Array<Object>> {
     const cachedValue = cache.get(query);
 
     const allStops = await AllStopUtils.fetchAllStops();
-    const formattedStops = allStops.map(s => ({
+    const filteredStops = allStops.filter(s => (
+      fuzz.partial_ratio(s.name.toLowerCase(), query.toLowerCase()) >= MIN_FUZZ_RATIO
+    ));
+    filteredStops.sort((a, b) => {
+      const aFuzzRatio = fuzz.partial_ratio(query, a.name.toLowerCase());
+      const bFuzzRatio = fuzz.partial_ratio(query, b.name.toLowerCase());
```
young-k: query is lowercased up above but not here, is there any reason? Also, if it's supposed to be lowercased, can we create a variable for it instead? maybe `normalizedQuery`

young-k: @kevinchan159 why did you resolve this conversation 😠, please enlighten me

mattbarker016:
It looks like this resolved #168 and #169, is there any way we can slide #173 (quick addition) into this as well?

kevinchan159:
> It looks like this resolves #168 and #169, is there any way we can slide #173 (quick addition) into this as well?

#173 will be done through the microservice so thats a different repo

PR #176, opened 2019-02-21
Clean up code, remove unnecessary checks

young-k:
Tried to simplify logic wherever possible, checked `logs.cornellappdev.com` to see if there were any Error cases that weren't being reached, verified that everything works on devserver.

### DIFF
--- a/src/routers/HelloWorldRouter.js
+++ b/src/routers/HelloWorldRouter.js
@@ -12,7 +12,7 @@ class HelloWorldRouter extends ApplicationRouter<string> {
   }
 
   // eslint-disable-next-line require-await
-  async content(req: Request) {
+  async content(req: Request): Promise<any> {
     return `Hello World! Environment: ${process.env.NODE_ENV || 'unknown'} | `
       + `Bus Navigation: http://${process.env.GHOPPER_BUS || 'ERROR'}:8988/ | `
       + `Walking Navigation: http://${process.env.GHOPPER_WALKING || 'ERROR'}:8987/ | `
--- a/src/routers/HelloWorldRouter.js
+++ b/src/routers/HelloWorldRouter.js
@@ -1,7 +1,6 @@
 // @flow
 import type Request from 'express';
 import ApplicationRouter from '../appdev/ApplicationRouter';
-import { init } from '../server';
 
 class HelloWorldRouter extends ApplicationRouter<string> {
   constructor() {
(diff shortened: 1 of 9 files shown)


src/routers/HelloWorldRouter.js
```diff
@@ -12,14 +11,12 @@ class HelloWorldRouter extends ApplicationRouter<string> {
     return '/';
   }
 
-  async content(req: Request): Promise<string> {
-    const checkInit = (req.query.awaitInit !== undefined && `${await init}.`) || process.env.PORT;
-
-    return (`Hello World! Environment: ${process.env.NODE_ENV || 'unknown'} | `
+  // eslint-disable-next-line require-await
```
cswenberg: don't know what this means but okay

src/utils/ParseRouteUtils.js
```diff
@@ -56,34 +56,28 @@ function createGpxJson(stops: Array<Object>, startTime: String): Object {
...
-      stops,
-      tripIdentifiers: tripIDs,
-    };
-  } catch (error) {
-    throw new Error(
-      LogUtils.logErr(error, { first, second }, 'mergeDirections failed'),
-    );
-  }
+  return {
+    type: first.type,
```
cswenberg: alphabetize?

src/utils/ParseRouteUtils.js
```diff
@@ -142,85 +136,79 @@ async function condenseRoute(
...
-  try {
-    // trim unnecessary walking directions
-    await trimFirstLastDirections(route, startCoords, endCoords);
+  // trim unnecessary walking directions
+  await trimFirstLastDirections(route, startCoords, endCoords);
+
+  let previousDirection = null;
+  let totalDistanceWalking = 0;
+
+  for (let index = 0; index < route.directions.length; index++) {
```
cswenberg: isn't forEach a thing for looping like this? you don't really use **index** for anything except the beginning

src/utils/ParseRouteUtils.js
```diff
@@ -142,85 +136,79 @@ async function condenseRoute(
...
+    const direction = route.directions[index];
+    const startTime = Date.parse(direction.startTime);
+    const endTime = Date.parse(direction.endTime);
+    /*
+     * Discard routes with directions that take over 2 hours time
```
cswenberg: one line comment?

src/utils/ParseRouteUtils.js
kevinchan159: classic

young-k: rip you got me, but i didn't even write this code fml

src/utils/ParseRouteUtils.js
```diff
@@ -142,85 +136,79 @@ async function condenseRoute(
...
+     */
+    if (startTime + (ONE_HOUR_IN_MILLISECONDS * 2) <= endTime) {
+      return null;
+    }
 
-    let previousDirection = null;
-    let totalDistanceWalking = 0;
+    /*
+     * Discard routes where not possible to walk to bus given departure buffer
```
cswenberg: one line comment?

src/routers/HelloWorldRouter.js
young-k: it's basically because it's a subclass and in the parent class it's defined as an `async` function, but in this specific instance we don't need the `async` functionality, if that makes any sense
so we ignore the linter error

src/utils/ParseRouteUtils.js
young-k: it's used on line 187, I didn't want to break existing functionality but I agree that this should be changed in the future, I'll make an issue for it!

webpack.config.js
```diff
@@ -7,34 +7,31 @@ console.log(path.resolve(''));
...
-            // What to watch.
-            watch: ['./build', './src/test', './src/server.js'],
-
-            // Detailed log.
-            verbose: false,
-        }),
-        new ExtraWatchWebpackPlugin({
-            dirs: ['./'],
-        }),
+  target: 'node',
```
cswenberg: alphabetize smh

young-k: i didnt frikin write the code

kevinchan159: @cswenberg you're a savage

PR #183, opened 2019-02-28
Fix travis deploy script to support multiple branches

meganle:
As the title says, the original syntax only recognized the first branch stated.

Just tested on this branch!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,9 +20,7 @@ deploy:
   provider: script
   script: bash docker_push
   on:
-    branches:
-      only:
-      - master
-      - release
+    all_branches: true
+    condition: $TRAVIS_BRANCH =~ ^master|release$
 script:
 - 


.travis.yml
```diff
@@ -20,9 +20,7 @@ deploy:
   provider: script
   script: bash docker_push
   on:
-    branches:
-      only:
-      - master
-      - release
+    all_branches: true
+    condition: $TRAVIS_BRANCH =~ ^master|release$
```
young-k: wait i'm actually confused, doesn't that contradict [this](https://docs.travis-ci.com/user/customizing-the-build/#safelisting-or-blocklisting-branches), i'm gonna ping you offline

.travis.yml
```diff
@@ -20,9 +20,7 @@ deploy:
...
 script:
```
young-k: is this tag required?

PR #184, opened 2019-02-28
Kevin/multiroute change

kevinchan159:
- `/multiroute` should `return the best available route for each destination specified from a single start location` but at the moment, it returns an array where each element is an array of routes for each destination. This PR fixes that.
- also added a command in `package.json` to start the backend locally so that we don't have to copy and paste the 3 commands every time

### DIFF
--- a/src/routers/MultiRouteRouter.js
+++ b/src/routers/MultiRouteRouter.js
@@ -1,5 +1,6 @@
 // @flow
 import type Request from 'express';
+import AnalyticsUtils from '../utils/AnalyticsUtils';
 import ApplicationRouter from '../appdev/ApplicationRouter';
 import LogUtils from '../utils/LogUtils';
 import RouteUtils from '../utils/RouteUtils';
@@ -34,8 +35,8 @@ class MultiRouteRouter extends ApplicationRouter<Array<Object>> {
       return [];
     }
 
-    // multiple destinations given
-    const routes = destinationNames.map((destinationName, index) => RouteUtils.getRoutes(
+    // get array of routes for each destination
+    const multiRoutes = destinationNames.map((destinationName, index) => RouteUtils.getRoutes(
       destinationName,
       end[index],
       start,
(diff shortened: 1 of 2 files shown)


src/routers/MultiRouteRouter.js
```diff
@@ -35,16 +36,21 @@ class MultiRouteRouter extends ApplicationRouter<Array<Object>> {
     }
 
     // multiple destinations given
```
yuna878: Comment should address that the variable is an array of routes to each destinations

src/routers/MultiRouteRouter.js
```diff
@@ -35,16 +36,21 @@ class MultiRouteRouter extends ApplicationRouter<Array<Object>> {
...
       departureTimeQuery,
       false,
     ));
 
-    return Promise.all(routes).then(val => val).catch((err) => {
-      throw LogUtils.logErr(err, routes, 'Could not get all specified routes');
+    // return the best route for each different destination
+    return Promise.all(multiRoutes).then((val) => {
+      const bestRoutes = val.map(routes => ((routes.length > 0) ? routes[0] : null));
+      AnalyticsUtils.assignRouteIdsAndCache(bestRoutes.filter(r => r !== null));
```
meganle: You can do `bestRoutes.filter(Boolean)` to filter out faulty values

package.json
```diff
@@ -13,6 +13,7 @@
     "start:dev": "./node_modules/.bin/webpack --watch --env.NODE_ENV=development --progress --colors --mode=development --debug --devtool cheap-eval-source-map --output-pathinfo",
     "start:microservices": "docker-compose up -d ghopper ghopper-walking live-tracking map",
     "start:node": "docker-compose up node",
+    "start:backend": "npm run build:dev; npm run start:microservices; npm run start:node",
```
young-k: you only have to run the first two commands one time, and it should hot reload & the microservices persist in the background, so i don't think we need this

src/routers/MultiRouteRouter.js
young-k: aren't non null checks `!=` and not `!==`?

young-k:
In the future, make sure to rename your PR titles as well

PR #188, opened 2019-03-05
Add 3 second timeouts to microservice requests

alanna-zhou:
Issue #187 & tested on dev server


src/routers/MultiRouteRouter.js
```diff
@@ -34,17 +35,22 @@ class MultiRouteRouter extends ApplicationRouter<Array<Object>> {
...
       start,
       departureTimeQuery,
       false,
     ));
 
-    return Promise.all(routes).then(val => val).catch((err) => {
-      throw LogUtils.logErr(err, routes, 'Could not get all specified routes');
+    // return the best route for each destination
+    return Promise.all(multiRoutes).then((val) => {
+      const bestRoutes = val.map(routes => ((routes.length > 0) ? routes[0] : null));
```
young-k: maybe just put the ternary on a separate line? 

```
const bestRoutes = val.map((routes) => {
    (routes.length > 0) ? routes[0] : null;
})

File: src/utils/RealtimeFeedUtils.js [hunk 2]
```diff
@@ -4,11 +4,14 @@ import LogUtils from './LogUtils';
...
 import TokenUtils from './TokenUtils';
 
+const THREE_SEC_IN_MS = 3000;
+
 async function fetchRTF(): Object {
   const options = {
     method: 'GET',
     url: `http://${PYTHON_APP || 'localhost'}:5000/rtf`,
     headers: { 'Cache-Control': 'no-cache' },
+    timeout: THREE_SEC_IN_MS,
```
young-k: can we put this in a `constants.js` similar to how pollo does it, since it's being referenced from a bunch of places, maybe like:

```
const GET_OPTIONS = { 
  method: 'GET',
  headers: { 'Cache-Control': 'no-cache' },
  timeout: THREE_SEC_IN_MS,
}
```

and then you can have:

```
const options = {
  ...GET_OPTIONS,
  url: `...`
}
```

src/routers/MultiRouteRouter.js
kq-li: ^ without the braces
```
const bestRoutes = val.map(routes => 
    routes.length > 0) ? routes[0] : null);
```

kq-li: though personally i prefer the original one liner

src/utils/AlertsUtils.js
```diff
@@ -2,11 +2,14 @@
 import { PYTHON_APP } from './EnvUtils';
 import RequestUtils from './RequestUtils';
 
+const THREE_SEC_IN_MS = 3000;
```
kq-li: rename to something about the function of the constant rather than what it physically is (`FETCH_TIMEOUT` or something similar)

src/routers/MultiRouteRouter.js
young-k: @meganle can you decide

src/utils/AlertsUtils.js
young-k: i feel like this is fine because in the code it's: `timeout: THREE_SEC_IN_MS`, but @meganle can you decide here as well

src/routers/MultiRouteRouter.js
meganle: I prefer the one liner, but do we need parentheses around `routes.length > 0`?

src/utils/AlertsUtils.js
meganle: i think the name is fine as is

src/utils/RealtimeFeedUtils.js
```diff
@@ -1,14 +1,15 @@
...
 import { PYTHON_APP } from './EnvUtils';
+import Constants from './Constants';
 import LogUtils from './LogUtils';
 import RequestUtils from './RequestUtils';
 import TokenUtils from './TokenUtils';
 
 async function fetchRTF(): Object {
   const options = {
-    method: 'GET',
+    ...Constants.GET_Options,
```
young-k: `GET_OPTIONS` here and everywhere else?

alanna-zhou: what do you mean

src/routers/MultiRouteRouter.js
```diff
@@ -46,7 +46,7 @@ class MultiRouteRouter extends ApplicationRouter<Array<Object>> {
 
     // return the best route for each destination
     return Promise.all(multiRoutes).then((val) => {
-      const bestRoutes = val.map(routes => ((routes.length > 0) ? routes[0] : null));
+      const bestRoutes = val.map(routes => (routes.length > 0 ? routes[0] : null));
```
young-k: you don't need the `()` around the ternary

young-k:
Also, can you make sure that none of the recent changes broke devserver?

src/routers/MultiRouteRouter.js
alanna-zhou: when i autosave, it gets added?

young-k: oh hm alright, @meganle i'm guessing it must be fine?

PR #192, opened 2019-03-09
Handle ghopper errors properly

kevinchan159:
Right now if ghopper fails to return any bus routes, we just throw an error. This makes the response something like this (https://transit-dev.cornellappdev.com/api/v1/route?arriveBy=0&destinationName=Current%20Location&end=42.43865970526117%2C-76.4841959719327&originName=Asbury%20%40%20Warren&start=42.524723%2C-76.472847&time=1552706100&uid=BFDA3D0E-1E38-4A66-ABBF-DAF1243B246B).

Instead of throwing an error, we should just log the error instead and continue to see if there are any walking routes.

### DIFF
--- a/src/routers/v1/RouteRouter.js
+++ b/src/routers/v1/RouteRouter.js
@@ -23,23 +23,25 @@ class RouteRouter extends ApplicationRouter<Array<Object>> {
       end,
       originName,
       start,
-      time: departureTimeQuery,
+      time,
       uid,
     } = params;
 
     const isArriveBy = (arriveBy === '1' || arriveBy === true);
-    const routes = await RouteUtils.getRoutes(destinationName, end, start, departureTimeQuery, isArriveBy);
-    const request = {
-      arriveBy,
-      destinationName,
-      end: routes[0].endCoords,
-      originName,
-      routeId: routes[0].routeId,
-      start: routes[0].startCoords,
-      time: departureTimeQuery,
-      uid,
-    };
-    LogUtils.log({ category: 'routeRequest', request });
+    const routes = await RouteUtils.getRoutes(destinationName, end, start, time, isArriveBy);
+    if (routes.length > 0) {
+      const request = {
+        arriveBy,
+        destinationName,
+        end: routes[0].endCoords,
+        originName,
+        routeId: routes[0].routeId,
+        start: routes[0].startCoords,
+        time,
+        uid,
+      };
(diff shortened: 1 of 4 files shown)


src/utils/RouteUtils.js
```diff
@@ -56,13 +56,14 @@ async function getRoutes(
   isArriveBy: boolean,
 ): Promise<Array<Object>> {
   const routeResponses = await GhopperUtils.fetchRoutes(end, start, departureTimeQuery, isArriveBy);
+  const { busRoute } = routeResponses;
```
alanna-zhou: can this be combined into one like megan's comment in your sectioned routes pr

kevinchan159: good catch. I changed walkingRoute to a const in that PR so once I merge that, I'll combine it into one line here

young-k: too speedy

src/routers/v1/RouteRouter.js
```diff
@@ -29,17 +29,19 @@ class RouteRouter extends ApplicationRouter<Array<Object>> {
...
-    LogUtils.log({ category: 'routeRequest', request });
+    if (routes.length > 0) {
+      const request = {
+        arriveBy,
+        destinationName,
+        end: routes[0].endCoords,
+        originName,
+        routeId: routes[0].routeId,
+        start: routes[0].startCoords,
+        time: departureTimeQuery,
```
young-k: i thought we were changing this to just `time`?

src/utils/RouteUtils.js
```diff
@@ -56,26 +56,26 @@ async function getRoutes(
...
+    return [];
   }
 
-  const { busRoute } = routeResponses;
-  let { walkingRoute } = routeResponses;
-  // parse the graphhopper walking route=
-  walkingRoute = ParseRouteUtils.parseWalkingRoute(
+  // parse the graphhopper walking route
+  const parsedWalkingRoute = ParseRouteUtils.parseWalkingRoute(
     walkingRoute,
```
young-k: can't `walkingRoute` be `null` at this point, since it won't be caught by the `if` statement on `61`? Is this acceptable behavior? Same thing regarding just `busRoute`, can't it be `null`? 

Should the `&&` be a `||` statement?

kevinchan159: it shouldn't be `||` b/c we only want to `return []` if we have no routes at all. i addressed it in a different way

Issue #190, opened 2019-03-09
Fix bus stop time estimates to accommodate delayed previous bus stops 


alanna-zhou:
From @chalo2000's thread in slack

young-k:
@AAAstorga this one's for you

PR #193, opened 2019-03-11
Minor refactoring; adjust walking speed and delay buffer

young-k:
Addresses #190 and makes a few more changes:

To summarize, let's say I was going from `A` to `B` and it would take me 5 minutes to walk to the bus stop, and the bus was departing at `10:00AM`. If GraphHopper believed that it would take 10 minutes for me to walk to the stop, the route wouldn't be shown at all. 

The fix was adjusting the delay buffer that was already present from 5 minutes to 20 minutes. I also verified that buses that leave before the `departAt` time are automatically excluded, so that bug won't be introduced either.

Screenshot of leaving at `11:37PM`, the bus is still shown:

IMG_06907802355F-1

- Spoke to @kevinchan159 offline about combining `RouteUtilsV2` and `RouteUtils` –– mainly we don't want to version anything that isn't a router, and I'll be refactoring `RouteUtils` to have more shared code in the future.
- Rename uses of `GhopperUtils` to `GraphhopperUtils`
- Renamed `delayBuffer` to `delayBufferMinutes`

I've tested on the devserver and verified that my changes work!


src/utils/RouteUtils.js
```diff
@@ -1,8 +1,162 @@
...
+ */
+async function getSectionedRoutes(
+  destinationName: string,
+  end: string,
+  start: string,
+  departureTimeQuery: number,
+  isArriveBy: boolean,
+  originBusStopName: ?string,
+): Promise<Object> {
+  const defaultState = {
```
kevinchan159: i don't know if `defaultState` is the best name here. maybe `defaultRoutes` or something

young-k: yeah i made the change to keep the types consistent

src/utils/ParseRouteUtils.js
```diff
@@ -151,13 +151,8 @@ async function condenseRoute(
...
 
     // Discard routes where not possible to walk to bus given departure buffer
-    if (departureDelayBuffer) {
-      if (direction.type === 'depart') {
-        const busActualDepartTime = startTime + (direction.delay != null ? direction.delay * 1000 : 0);
-        if (busActualDepartTime < departureTimeNowMs) {
-          return null;
-        }
-      }
+    if (departureDelayBuffer && direction.type === 'depart' && startTime < departureTimeNowMs) {
```
young-k: I'm not sure why we factor in a `delay` and adjust the departTime for that, and I encountered bugs where it was messing up calculations, so I removed it!

Issue #195, opened 2019-03-11
Refactor ParseRouteUtils


young-k:
This file has always been difficult to understand, I'm going to rewrite it and add documentation

PR #196, opened 2019-03-11
Refactor RouteUtils

young-k:
First of a few changes addressing #195.

List of changes made:

- Split shared code from `createSectionedRoute` and `getRoutes` into a helper function.
- There's an invariant that a `walkingRoute` will always be returned, I changed our code to anticipate that.
- `adjustRouteTimesIfNecessary` was being used in `RouteUtils`, this should be private to `ParseRouteUtils` to make the logic cleaner.

Screenshot verifying it works on devserver:

IMG_CA1F295A32AA-1

### DIFF
--- a/src/appdev/ApplicationRouter.js
+++ b/src/appdev/ApplicationRouter.js
@@ -102,7 +102,7 @@ class ApplicationRouter<T> {
     return async (req: Request, res: Response, next: NextFunction) => {
       try {
         const content = await this.content(req);
-        LogUtils.log({ path: this.getPath(), query: req.query, response: content });
+        LogUtils.log({ path: this.getPath(), query: req.query });
         res.json(new AppDevResponse(true, content));
       } catch (e) {
         if (e.message === 1) {
--- a/src/utils/RouteUtils.js
+++ b/src/utils/RouteUtils.js
@@ -90,6 +90,38 @@ async function createSectionedRoutes(
 }
 
 /**
+ * Queries Graphhopper and returns exactly one walking route and any bus routes.
+ *
+ * The routes are processed prior to being returned. Note that GraphHopper always returns a walking route.
+ *
+ * @param destinationName
+ * @param end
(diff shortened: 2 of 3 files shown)


src/appdev/ApplicationRouter.js
```diff
@@ -102,7 +102,7 @@ class ApplicationRouter<T> {
     return async (req: Request, res: Response, next: NextFunction) => {
       try {
         const content = await this.content(req);
-        LogUtils.log({ path: this.getPath(), query: req.query, response: content });
+        LogUtils.log({ path: this.getPath(), query: req.query });
```
kevinchan159: why aren't we logging `content` anymore?

src/utils/RouteUtils.js
```diff
@@ -89,6 +89,38 @@ async function createSectionedRoutes(
...
+  isArriveBy: boolean,
+): Promise<{ parsedBusRoutes: ?Array<Object>, parsedWalkingRoute: Object }> {
+  const { busRoutes, walkingRoute } = await GraphhopperUtils.fetchRoutes(end, start, departureTimeQuery, isArriveBy);
+  const parsedWalkingRoute = ParseRouteUtils.parseWalkingRoute(
+    walkingRoute,
+    GraphhopperUtils.getDepartureTime(departureTimeQuery, isArriveBy, 0),
+    destinationName,
+  );
+
+  if (!busRoutes) return { parsedWalkingRoute, parsedBusRoutes: busRoutes };
```
kevinchan159: would it be clearer to just `return { parsedWalkingRoute, parsedBusRoutes: null }`

src/appdev/ApplicationRouter.js
young-k: it's huge, and it was more important when we were trying to debug errors – i think even though we should be logging any time a client makes a request, logging stuff like live tracking responses isn't important

PR #197, opened 2019-03-11
Fix bug where buses leaving at current minute did not display

young-k:
It's fixed:

IMG_2689

Verifying that 6:54 doesn't show the 6:53 bus:

IMG_2690

I also added `.git` to `.dockerignore` because we don't need it in the build!

### DIFF
--- a/src/utils/ParseRouteUtils.js
+++ b/src/utils/ParseRouteUtils.js
@@ -8,8 +8,8 @@ import LogUtils from './LogUtils';
 import RealtimeFeedUtils from './RealtimeFeedUtils';
 import RequestUtils from './RequestUtils';
 
-const ONE_HOUR_IN_MILLISECONDS = 3600000;
-const ONE_MIN_IN_MILLISECONDS = 60000;
+const ONE_HOUR_IN_MS = 3600000;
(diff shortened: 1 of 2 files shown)


src/utils/ParseRouteUtils.js
```diff
@@ -150,8 +150,14 @@ async function condenseRoute(
...
-    // Discard routes where not possible to walk to bus given departure buffer
-    if (departureDelayBuffer && direction.type === 'depart' && startTime < departureTimeNowMs) {
+    /**
+     * Discard routes where not possible to walk to bus given departure buffer
+     * We subtract (ONE_MIN_IN_MILLISECONDS - 1) from departureTimeNow to avoid the
+     * case where we don't show a bus leaving at the same minute as departureTime.
+     */
+    if (departureDelayBuffer
+      && direction.type === 'depart'
+      && startTime < departureTimeNowMs - (ONE_MIN_IN_MILLISECONDS - 1)) {
```
alanna-zhou: that's kind of a long name.. why not just ONE_MIN_IN_MS

young-k: will change, i did not write this file 😠

PR #198, opened 2019-03-12
Refactor RouteUtils further

young-k:
Closer to finishing #195!

This file is also flow-error free now, and I ended up removing 55 LOC :)

### DIFF
--- a/src/utils/RouteUtils.js
+++ b/src/utils/RouteUtils.js
@@ -24,11 +24,10 @@ async function isBusStop(location: string): Promise<boolean> {
 }
 
 /**
- * Filter and validate the array of parsed routes to send to the client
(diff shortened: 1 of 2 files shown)


src/utils/RouteUtils.js
```diff
@@ -145,73 +120,40 @@ async function getSectionedRoutes(
...
-      startPoint,
-      endPoint,
-      walkingRoute.directions[0].distance,
-      departureDelayBuffer,
-      departureTimeNowMs,
-    )),
-  )).filter(route => route !== null);
+  finalBusRoutes.forEach((route) => {
+    if (originBusStopName !== null) {
+      if (route.directions && route.directions.length > 0) {
```
yuna878: Can you combine lines 144 and 145 because there's no else body for both?

PR #202, opened 2019-03-13
Add epoch and formatted time to logs

alanna-zhou:
Add epoch and formatted EST time to logging requests

### DIFF
--- a/src/appdev/ApplicationRouter.js
+++ b/src/appdev/ApplicationRouter.js
@@ -2,7 +2,7 @@
 import {
   NextFunction, Request, Response, Router,
 } from 'express';
-import AppDevUtilities from './ApplicationUtils';
+import ApplicationUtils from './ApplicationUtils';
 import LogUtils from '../utils/LogUtils';
 
 /**
@@ -57,7 +57,7 @@ class ApplicationRouter<T> {
     const path = this.getPath();
 
     // Make sure path conforms to specification
-    AppDevUtilities.tryCheckAppDevURL(path);
+    ApplicationUtils.tryCheckAppDevURL(path);
 
     // Attach content to router
     this.requestTypes.forEach((reqType) => {
@@ -95,6 +95,13 @@ class ApplicationRouter<T> {
   }
 
   /**
+   * Gets a formatted string of the date and time right now in the EST time zone.
+   */
+  _getESTDateStringNow() {
+    return new Date().toLocaleString('en-US', { timeZone: 'America/New_York' });
+  }
+
+  /**
    * Create a wrapper around the response hook to pass to the Express
    * Router.
    */
@@ -102,7 +109,12 @@ class ApplicationRouter<T> {
     return async (req: Request, res: Response, next: NextFunction) => {
       try {
         const content = await this.content(req);
-        LogUtils.log({ path: this.getPath(), query: req.query });
+        LogUtils.log({
+          path: this.getPath(),
+          query: req.query,
+          serverEpochTime: Date.now(),
+          serverFormattedTime: this._getESTDateStringNow(),
+        });
         res.json(new AppDevResponse(true, content));
       } catch (e) {
         if (e.message === 1) {
--- a/src/appdev/ApplicationUtils.js
+++ b/src/appdev/ApplicationUtils.js
@@ -17,22 +17,6 @@ const tryCheckAppDevURL = (path: string) => {
   }
 };
 
-const insertIntoMySQLStatement = (
-  tableName: string,
-  fields: Object,
-): string => {
-  const columns = `(${Object.keys(fields).join(', ')})`;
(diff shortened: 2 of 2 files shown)


src/appdev/ApplicationRouter.js
```diff
@@ -102,7 +102,9 @@ class ApplicationRouter<T> {
     return async (req: Request, res: Response, next: NextFunction) => {
       try {
         const content = await this.content(req);
-        LogUtils.log({ path: this.getPath(), query: req.query });
+        LogUtils.log({
+          path: this.getPath(), query: req.query, epochTime: Date.now(), formattedTime: new Date(Date.now()).toUTCString(),
```
kevinchan159: make sure to alphabetize after you fix the hound issue

src/appdev/ApplicationRouter.js
```diff
@@ -102,7 +102,12 @@ class ApplicationRouter<T> {
     return async (req: Request, res: Response, next: NextFunction) => {
       try {
         const content = await this.content(req);
-        LogUtils.log({ path: this.getPath(), query: req.query });
+        const epochTime = Date.now();
+        const formattedTime = new Date(epochTime).toUTCString();
+        const message = {
+          epochTime, formattedTime, path: this.getPath(), query: req.query,
```
young-k: maybe `serverEpochTime`, `serverFormattedTime` just to be extra clear

src/appdev/ApplicationRouter.js
```diff
@@ -102,7 +102,12 @@ class ApplicationRouter<T> {
...
+        };
+        LogUtils.log(message);
```
young-k: why not:

```
LogUtils.log({ 
  path: this.getPath(),
  query: req.query,
  ...
});
```

I don't think there's a point of declaring a variable just to use it in the next line!

src/appdev/ApplicationUtils.js
```diff
@@ -32,7 +32,14 @@ const insertIntoMySQLStatement = (
   return `INSERT INTO ${tableName} ${columns} VALUES ${values};`;
 };
 
+function getESTDateString() {
+  let usaTime = new Date().toLocaleString('en-US', { timeZone: 'America/New_York' });
```
young-k: this is estTime not usatime

src/appdev/ApplicationUtils.js
```diff
@@ -32,7 +32,14 @@ const insertIntoMySQLStatement = (
...
+  usaTime = new Date(usaTime);
+  return usaTime.toLocaleString();
```
young-k: wait i don't understand why you have to do this twice? you make the date, make it a string, make it a date, and make it a string again?

alanna-zhou: the first line gets the string, and the second line creates a date object

young-k: right but why do you need lines 2 and 3? the first one makes the date and makes it a string!

src/appdev/ApplicationRouter.js
```diff
@@ -102,7 +102,12 @@ class ApplicationRouter<T> {
     return async (req: Request, res: Response, next: NextFunction) => {
       try {
         const content = await this.content(req);
-        LogUtils.log({ path: this.getPath(), query: req.query });
+        LogUtils.log({
+          path: this.getPath(),
+          query: req.query,
+          serverEpochTime: Date.now(),
+          serverFormattedTime: AppDevUtilities.getESTDateString(),
```
young-k: this helper function should probably be just defined here, unless other files are going to use it in the future! Also `AppDevUtilities` should be `ApplicationUtils`, we want to keep the files/imports consistent

young-k: you can just use the `_` convention that Pollo uses, for private functions in a class!

alanna-zhou: utils should be helper methods though right? applicationutils is just a file that holds helper functions for applicationrouter

young-k: also can we add javadoc, and change the name to `getESTDateStringNow`

alanna-zhou:
> Address comments, else LGTM. Did you verify that the times are in EST? We want this to be as clear to us as possible!

Yes!

src/appdev/ApplicationRouter.js
young-k: there's only one other function that's being used in `ApplicationUtils`, can we just deprecate the file completely and move these two functions here? 

cc: @meganle thoughts

meganle: well the other function in `ApplicationUtils` should stay since that verifies all of the urls in all of the routers. i think it's fine to add helper methods to that file. i know pollo uses it for misc helper functions so it should be fine there

Issue #213, opened 2019-03-23
Show delayed routes on route display page


kevinchan159:
- Right now, if it is past 11:15pm, we don't see the first bus route even though we should since its delayed

![Image from iOS](https://user-images.githubusercontent.com/26048121/54861112-55692580-4cfa-11e9-8589-0d18f208dae8.jpg)

young-k:
closed by #218

PR #218, opened 2019-03-26
Make 2 Ghopper requests to display delayed routes

kevinchan159:
Addresses #213 

Screenshot:
55606207_400350694079911_864107270600916992_n
55508157_477347016135618_4856002144978862080_n

### DIFF
--- a/src/utils/GraphhopperUtils.js
+++ b/src/utils/GraphhopperUtils.js
@@ -5,7 +5,7 @@ import {
 import LogUtils from './LogUtils';
 import RequestUtils from './RequestUtils';
 
-const DELAY_BUFFER_IN_MINUTES = 8; // buffer for fast walking speeds
+const DELAY_BUFFER_IN_MINUTES = 20; // buffer for fast walking speeds
 
 /**
  * https://graphhopper.com/api/1/docs/routing/#output
@@ -14,10 +14,16 @@ const DELAY_BUFFER_IN_MINUTES = 8; // buffer for fast walking speeds
  * @param departureTimeQuery
  * @param arriveBy
  */
-const getGraphhopperBusParams = (end: string, start: string, departureTimeQuery: string, arriveBy: boolean) => ({
+const getGraphhopperBusParams = (
+  end: string,
+  start: string,
+  departureTimeQuery: string,
+  arriveBy: boolean,
+  delayBufferMinutes: number,
+) => ({
   'ch.disable': true,
   'pt.arrive_by': arriveBy,
-  'pt.earliest_departure_time': getDepartureTimeDateNow(departureTimeQuery, arriveBy, DELAY_BUFFER_IN_MINUTES),
+  'pt.earliest_departure_time': getDepartureTimeDate(departureTimeQuery, arriveBy, delayBufferMinutes),
   'pt.max_walk_distance_per_leg': 2000,
   'pt.profile': true,
   'pt.walk_speed': 3.0, // > 3.0 suggests getting off bus earlier and walk half a mile instead of waiting longer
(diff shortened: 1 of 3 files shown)


src/utils/GraphhopperUtils.js
```diff
@@ -121,22 +155,39 @@ async function fetchRoutes(end: string, start: string, departureTimeDateNow: str
...
+  // Fetch buses using the current start time
+  const busOptionsNow = {
+    qs: getGraphhopperBusParams(end, start, departureTimeDateNow, isArriveByQuery, 0),
+    url: `http://${GHOPPER_BUS || 'ERROR'}:8988/route`,
+    ...sharedOptions,
+  };
+
+  // Fetch buses using a delay buffer of DELAY_BUFFER_IN_MINUTES.
+  // This allows us to display routes that are delayed
+  const busOptionsBuffered = {
```
young-k: maybe `busOptionsWithDelay` would be a better name for this?

src/utils/GraphhopperUtils.js
```diff
@@ -148,15 +199,32 @@ async function fetchRoutes(end: string, start: string, departureTimeDateNow: str
...
+      busRouteNowRequest && busRouteNowRequest.body,
+      getGraphhopperBusParams(end, start, departureTimeDateNow, isArriveByQuery),
+      `Routing failed: ${GHOPPER_BUS || 'undefined graphhopper bus env'}`,
+    );
+  }
+
+  if (busRouteBufferedRequest && busRouteBufferedRequest.statusCode < 300) {
+    const bufferedBusRoutes = JSON.parse(busRouteBufferedRequest.body).paths;
+    const validBusRoutes = [];
+    bufferedBusRoutes.forEach((bufferedRoute) => {
```
young-k: why not just `filter`, instead of a `forEach`?

src/utils/ParseRouteUtils.js
```diff
@@ -150,11 +150,14 @@ async function condenseRoute(
...
      * case where we don't show a bus leaving at the same minute as departureTime.
+     * We also have to account for the delay time because if a bus is delayed, we want to display
+     * the route even if the start time is earlier than the current time.
      */
-    if (departureDelayBuffer
-      && direction.type === 'depart'
-      && startTime < departureTimeNowMs - (ONE_MIN_IN_MS - 1)) {
-      return null;
+    if (departureDelayBuffer && direction.type === 'depart') {
+      const delayMs = (direction.delay !== null) ? direction.delay * 1000 : 0;
```
meganle: i don't think you need parentheses here

src/utils/ParseRouteUtils.js
```diff
@@ -150,11 +150,14 @@ async function condenseRoute(
...
+      if (startTime < departureTimeNowMs - (ONE_MIN_IN_MS - 1) - delayMs) {
```
meganle: same here, unless you're using parentheses to make it clearer

kevinchan159: i think it makes it clearer so i'll keep it

src/utils/GraphhopperUtils.js
```diff
@@ -121,22 +166,41 @@ async function fetchRoutes(end: string, start: string, departureTimeDateNow: str
...
+  // Fetch bus routes using the current start time
+  const busOptionsNow = {
+    qs: getGraphhopperBusParams(end, start, departureTimeDateNow, isArriveByQuery, 0),
     url: `http://${GHOPPER_BUS || 'ERROR'}:8988/route`,
     ...sharedOptions,
   };
+
+  // Fetch bus routes using a delay buffer of DELAY_BUFFER_IN_MINUTES.
+  // This means that we are fetching routes with the start time as DELAY_BUFFER_IN_MINUTES
+  // ago from the current start time.
```
young-k: Maybe  just`Fetch routes with (startTime - DELAY_BUFFER_IN_MINUTES)`, and that's more clear?

src/utils/GraphhopperUtils.js
```diff
@@ -148,15 +212,31 @@ async function fetchRoutes(end: string, start: string, departureTimeDateNow: str
       true,
     ),
   ]).then((vals) => {
```
young-k: why not just `[busRouteNowRequest, busRouteBufferedRequest, walkingRouteRequest] = await ...`
we shouldn't be using `.then()` with `Promise`

src/utils/GraphhopperUtils.js
```diff
@@ -148,15 +212,31 @@ async function fetchRoutes(end: string, start: string, departureTimeDateNow: str
...
+    LogUtils.log(
+      busRouteNowRequest && busRouteNowRequest.body,
+      getGraphhopperBusParams(end, start, departureTimeDateNow, isArriveByQuery),
+      `Routing failed: ${GHOPPER_BUS || 'undefined graphhopper bus env'}`,
+    );
+  }
+
+  if (busRouteBufferedRequest && busRouteBufferedRequest.statusCode < 300) {
+    const bufferedBusRoutes = JSON.parse(busRouteBufferedRequest.body).paths;
+    const validBusRoutes = bufferedBusRoutes.filter((bufferedRoute) => {
```
young-k: beautiful variable naming i am a fan

src/utils/GraphhopperUtils.js
```diff
@@ -148,15 +212,31 @@ async function fetchRoutes(end: string, start: string, departureTimeDateNow: str
...
+      const isDuplicateRoute = busRoutes.find(busRoute => busRoutesAreEqual(bufferedRoute, busRoute)) !== undefined;
```
young-k: probably cleaner to do:

`busRoutes.find(busRoute => busRoutesAreEqual(bufferedRoute, busRoute)) || false`;

src/utils/ParseRouteUtils.js
```diff
@@ -150,11 +150,14 @@ async function condenseRoute(
...
      * case where we don't show a bus leaving at the same minute as departureTime.
+     * We also have to account for the delay time because if a bus is delayed, we want to display
+     * the route even if the start time is earlier than the current time.
      */
-    if (departureDelayBuffer
-      && direction.type === 'depart'
-      && startTime < departureTimeNowMs - (ONE_MIN_IN_MS - 1)) {
-      return null;
+    if (departureDelayBuffer && direction.type === 'depart') {
+      const delayMs = direction.delay !== null ? direction.delay * 1000 : 0;
```
young-k: Can `direction.delay` really be null? I'm not sure if this check is necessary

src/utils/ParseRouteUtils.js
```diff
@@ -150,11 +150,14 @@ async function condenseRoute(
...
+      if (startTime < departureTimeNowMs - (ONE_MIN_IN_MS - 1) - delayMs) {
```
young-k: Do we need to subtract `1`, since we have `<` and not `<=`? If we don't, we should drop the parentheses

src/utils/GraphhopperUtils.js
kevinchan159: wouldn't that find the `busRoutes` that are duplicates?

src/utils/ParseRouteUtils.js
kevinchan159: yeah it can, i've seen it in responses

src/utils/GraphhopperUtils.js
```diff
@@ -121,22 +165,40 @@ async function fetchRoutes(end: string, start: string, departureTimeDateNow: str
...
     url: `http://${GHOPPER_WALKING || 'ERROR'}:8987/route`,
     ...sharedOptions,
   };
 
-  let busRouteRequest;
+  let busRouteNowRequest;
+  let busRouteBufferedRequest;
   let walkingRouteRequest;
-  await Promise.all([
+  [busRouteNowRequest, busRouteBufferedRequest, walkingRouteRequest] = await Promise.all([
```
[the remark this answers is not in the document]
kevinchan159: if i change them to `const`, it won't work b/c you have to initialize `const` with a value which means i can't assign them in this line

PR #219, opened 2019-03-26
Remove delays that are less than two minutes 

alanna-zhou:
### **before:**
issue217beforepic

### **after:**
issue217afterpic

### DIFF
--- a/src/utils/RealtimeFeedUtils.js
+++ b/src/utils/RealtimeFeedUtils.js
@@ -165,7 +165,7 @@ function getDelayInformation(stopID: String, tripID: String, rtf: Object): ?Obje
   const info = rtf[tripID];
   let delay = parseInt(info.stopUpdates && info.stopUpdates[stopID]);
   if (Number.isNaN(delay)) delay = parseInt(info.delay);
-
+  if (delay <= 120) delay = 0;
   return {
     delay,
     vehicleId: parseInt(info.vehicleId),


src/utils/RealtimeFeedUtils.js
```diff
@@ -165,7 +165,7 @@ function getDelayInformation(stopID: String, tripID: String, rtf: Object): ?Obje
   const info = rtf[tripID];
   let delay = parseInt(info.stopUpdates && info.stopUpdates[stopID]);
   if (Number.isNaN(delay)) delay = parseInt(info.delay);
-
+  if (delay <= 120) delay = 0;
   return {
     delay,
```
meganle: you could remove the if statement above and just do `delay <= 120 ? 0 : delay` here

young-k: wait @meganle i think we should keep it, and add a big comment directly above as to why we're doing this –– else it will be unclear to people reading in the future

meganle: yeah that works! or we can make `120` into a constant called like `MIN_DELAY_BUFFER` or something more descriptive or similar since 120 is very arbitrary?

young-k:
@alanna-zhou wait there's a bigger issue:

Let's say it's 4:00PM and I check routes from A to B, but there's a bus at 3:58PM that is delayed by 2 minutes. This bus **should display the delay**, or else the user won't understand why we're displaying that bus at 4:00PM. Can you make sure that this edge case is covered, and attach a screenshot when you do?

alanna-zhou:
> @alanna-zhou wait there's a bigger issue:
> 
> Let's say it's 4:00PM and I check routes from A to B, but there's a bus at 3:58PM that is delayed by 2 minutes. This bus **should display the delay**, or else the user won't understand why we're displaying that bus at 4:00PM. Can you make sure that this edge case is covered, and attach a screenshot when you do?

young-k:
It needs to display 4:00 though, if it shows 3:58PM then the user won't understand why we're showing a bus that they needed to board 2 minutes ago!

PR #223, opened 2019-03-27
Log requests that contain a route with transfer

kevinchan159:
Addresses #144 

I added a `containsTransfer` field to the LogUtils in `RouteRouter` which will tell us if the current route request returned a route w/ a transfer.

### DIFF
--- a/src/routers/v1/RouteRouter.js
+++ b/src/routers/v1/RouteRouter.js
@@ -29,7 +29,8 @@ class RouteRouter extends ApplicationRouter<Array<Object>> {
 
     const isArriveBy = (arriveBy === '1' || arriveBy === true);
     const routes = await RouteUtils.getRoutes(destinationName, end, start, time, isArriveBy);
-    if (routes.length > 0) {
+    const containsTransfer = routes.find(route => RouteUtils.routeContainsTransfer(route)) !== undefined;
+    if (routes.length > 0 && containsTransfer) {
       const request = {
         arriveBy,
         destinationName,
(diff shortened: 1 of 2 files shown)


src/utils/RouteUtils.js
```diff
@@ -25,6 +25,22 @@ async function isBusStop(location: string): Promise<boolean> {
...
+ *
+ * @param route
+ * @returns {boolean}
+ */
+function routeContainsTransfer(route: Object): boolean {
+  const { directions } = route;
+  const routeNumbers = [];
+  directions.forEach((direction) => {
+    const { routeNumber } = direction;
+    if (routeNumber !== null && !routeNumbers.includes(routeNumber)) routeNumbers.push(routeNumber);
```
meganle: could you just do `if (routeNumber &&...`

src/routers/v1/RouteRouter.js
```diff
@@ -40,7 +40,8 @@ class RouteRouter extends ApplicationRouter<Array<Object>> {
         time,
         uid,
       };
-      LogUtils.log({ category: 'routeRequest', request });
+      const containsTransfer = routes.find(route => RouteUtils.routeContainsTransfer(route)) !== undefined;
+      LogUtils.log({ category: 'routeRequest', request, containsTransfer });
```
young-k: we should only log if it actually contains a transfer, not if it doesn't (the majority of routes don't contain transfers)

kevinchan159: well right now we log every request anyways so I just added a field to that log

young-k: discussed offline, we should remove this log

PR #230, opened 2019-04-18
Add fields to Direction in swagger

kevinchan159:


kevinchan159:
> wait i thought this wasn't allowed in swagger.
> from my experience i've had to do type: number, format: double for swagger to avoid errors? could you confirm this

it seems fine i think:
Screen Shot 2019-04-18 at 10 42 05 AM

alanna-zhou:
> > wait i thought this wasn't allowed in swagger.
> > from my experience i've had to do type: number, format: double for swagger to avoid errors? could you confirm this
> 
> it seems fine i think:
> Screen Shot 2019-04-18 at 10 42 05 AM

kevinchan159:
OH wait i see. yah there was an error when i moved it to swagger editor. ill fix it, nice catch

PR #233, opened 2019-04-27
Add sorting for bus routes

kevinchan159:
Right now we don't do any sort of sorting for routes which leads to this issue sometimes where the first route is objectively worse than the second route.
58383285_344407272747759_1276420402985304064_n

Fixed:
Screen Shot 2019-04-27 at 7 07 38 PM

### DIFF
--- a/src/utils/RouteUtils.js
+++ b/src/utils/RouteUtils.js
@@ -42,6 +42,27 @@ function routeContainsTransfer(route: Object): boolean {
 }
 
 /**
+ * Returns whether [routeA] and [routeB] share the the same start and end bus stops.
+ * @param routeA
+ * @param routeB
+ * @returns {boolean}
+ */
+function routesHaveSameStartEndStops(routeA: Object, routeB: Object): boolean {
+  const routeADirections = routeA.directions;
+  const routeBDirections = routeB.directions;
+
+  if (routeADirections.length < 2 || routeBDirections.length < 2) {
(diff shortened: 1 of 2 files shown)


src/utils/RouteUtils.js
```diff
@@ -80,7 +101,17 @@ async function createFinalBusRoutes(
...
       departureTimeNowMs,
     )),
-  )).filter(route => route !== null);
+  )).filter(route => route !== null).sort((routeA, routeB) => {
+    // For routes that have the same start and end bus stops, the route with
+    // the earlier departure time should be shown first.
+    if (routesHaveSameStartEndStops(routeA, routeB)) {
+      const routeADepartureTime = new Date(routeA.departureTime);
+      const routeBDepartureTime = new Date(routeB.departureTime);
+      return (routeADepartureTime < routeBDepartureTime) ? -1 : 1;
```
meganle: i don't think you need parens around this

src/utils/RouteUtils.js
```diff
@@ -41,6 +41,27 @@ function routeContainsTransfer(route: Object): boolean {
   return routeNumbers.length > 1;
 }
 
+/**
+ * Returns whether [routeA] and [routeB] share the the same start and end bus stops.
+ * @param {*} routeA
```
yuna878: Delete `{*}`!

Issue #41, opened 2018-08-29
v2: Refactor Models


mattbarker016:
# Models

## Route
- departureTime: Date
- arrivalTime: Date
- startCoords: Coordinates // see below
- endCoords: Coordinates // see below
- travelDistance: Double // distance between start and end coords in miles
- startName: String
- endName: String
- boundingBox: Bounds // see below
- totalDuration: Int // minutes between departureTime and arrivalTime
- routeSummary: [RouteSummaryElement] // see @myo3's comment below
- detailDirections: [Direction] // see comment below

## Direction
- type: DirectionType
- name: String // General Description (bus stop name, Google Place name)
- startLocation: Location
- endLocation: Location
- startTime: Date
- endTime: Date
- path: [Coordinates]
- travelDistance: Double
- routeNumber: Int
- stops: [Location]
- stayOnBusTransfer: Bool
- tripIdentifiers: [String]?
- delay: Int?

### detailDirections Logic
*see Route.swift Line 164 on for front-end implantation*

1. You will need to create `arrive` and `.transfer` directions. Use `DirectionType` accordingly and *correctly*. `stayOnBusTransfer` will help with the latter.
2. Remove the first and last stop in each direction with stops if stops isn't empty.
3. Make sure the last direction's name is the ending location.

A typical route should be
```
[.walk, .depart, .arrive, .walk]
```

A route with two buses should be
```
[.walk, .depart, .arrive, .depart, .arrive, .walk]
```
There is no walking at `index 3` because the user waits at the bus stop.

A route with a stay on bus transfer should be
```
[.walk, .depart, .transfer, .arrive, .walk]
```

# Other Relevant Classes

**Bounds** 
```
"bounds" :  {
    "minLat" : Double,
    "minLong" : Double,
    "maxLat" : Double,
    "maxLong" : Double
}
```

**Coordinates**
```
"coordinates" : {
    "lat" : Double,
    "long" : Double
}
```

**DirectionType**
```
enum DirectionType: String {
    case walk // Directions that involving walking
    case depart // Directions where the user gets on the bus
    case arrive // Direction where the user gets off the bus
    case transfer // IMPORTANT: Direction where transfer is involved, but user stays on bus
}
```

**Location** 
```
"location" :  {
    "name" : String,
    "id" : String, // Identifier associated with location, used mainly for stopID for bus stop locations
    "lat" : Double,
    "long" : Double
}
```

myo3:
# Route Summary
[...]

PR #234, opened 2019-04-29
Kevin/refactor models

kevinchan159:
Addresses #41 
- Refactoring our models so that client side will require less logic.
- Basically adds `travelDistance, startName, endName, totalDuration, routeSummary` fields to every `Route` object.
- Since we are only adding fields, this shouldn't break any old logic on client

### DIFF
--- a/src/routers/v1/MultiRouteRouter.js
+++ b/src/routers/v1/MultiRouteRouter.js
@@ -5,6 +5,8 @@ import ApplicationRouter from '../../appdev/ApplicationRouter';
 import LogUtils from '../../utils/LogUtils';
 import RouteUtils from '../../utils/RouteUtils';
 
+const CURRENT_LOCATION = 'Current Location';
+
 /**
  * Router object that returns an array of the best available route for each
  * destination, specified from a single start destination.
@@ -37,6 +39,7 @@ class MultiRouteRouter extends ApplicationRouter<Array<Object>> {
 
     // get array of routes for each destination
     const multiRoutes = destinationNames.map((destinationName, index) => RouteUtils.getRoutes(
+      CURRENT_LOCATION,
       destinationName,
       end[index],
       start,
--- a/src/utils/ParseRouteUtils.js
+++ b/src/utils/ParseRouteUtils.js
@@ -8,6 +8,7 @@ import LogUtils from './LogUtils';
 import RealtimeFeedUtils from './RealtimeFeedUtils';
 import RequestUtils from './RequestUtils';
 
+const ONE_DAY_IN_MS = 86400000;
 const ONE_HOUR_IN_MS = 3600000;
 const ONE_MIN_IN_MS = 60000;
 
@@ -262,6 +263,49 @@ function generateBoundingBox(path: Object): Object {
 }
 
 /**
+ * Returns the distance between [startCoords] and [endCoords] in miles
+ * @param startCoords
+ * @param endCoords
+ * @returns {number}
+ */
+function getDistanceInMiles(startCoords: Object, endCoords: Object): number {
+  const lat1 = startCoords.lat;
+  const lon1 = startCoords.long;
+  const lat2 = endCoords.lat;
+  const lon2 = endCoords.long;
+  if (lat1 === lat2 && lon1 === lon2) {
+    return 0;
+  }
+
+  const radlat1 = Math.PI * lat1 / 180;
+  const radlat2 = Math.PI * lat2 / 180;
+  const theta = lon1 - lon2;
+  const radtheta = Math.PI * theta / 180;
+  let dist = Math.sin(radlat1) * Math.sin(radlat2) + Math.cos(radlat1) * Math.cos(radlat2) * Math.cos(radtheta);
+  if (dist > 1) {
+    dist = 1;
+  }
+  dist = Math.acos(dist);
+  dist = dist * 180 / Math.PI;
+  dist = dist * 60 * 1.1515;
+  return dist;
+}
+
+/**
+ * Returns the difference between [departureTime] and [arrivalTime] in minutes.
+ * @param departureTime
+ * @param arrivalTime
+ * @returns {number}
+ */
+function getDifferenceInMinutes(departureTime: string, arrivalTime: string): number {
+  const departureDate = new Date(departureTime);
+  const arrivalDate = new Date(arrivalTime);
+  const diffInMs = arrivalDate - departureDate;
+  const diffInMins = Math.round(((diffInMs % ONE_HOUR_IN_MS) % ONE_HOUR_IN_MS) / ONE_MIN_IN_MS);
+  return diffInMins;
+}
+
+/**
  * Returns dateInMs in an ISO String
  *
  * @param dateInMs
@@ -357,6 +401,7 @@ function parseBusRoutes(busRoutes: Array<Object>, destinationName: string): Prom
       // string 2018-02-21T17:27:00Z
       const { departureTime } = legs[0];
       const arriveTime = legs[numberOfLegs - 1].arrivalTime;
+      const totalDuration = getDifferenceInMinutes(departureTime, arriveTime);
 
       const startingLocationGeometry = legs[0].geometry;
       const endingLocationGeometry = legs[numberOfLegs - 1].geometry;
@@ -365,6 +410,7 @@ function parseBusRoutes(busRoutes: Array<Object>, destinationName: string): Prom
         startCoords,
         endCoords,
       } = getStartEndCoords(startingLocationGeometry, endingLocationGeometry);
+      const travelDistance = getDistanceInMiles(startCoords, endCoords);
       const boundingBox = generateBoundingBox(busRoute);
 
       const directions = await Promise.all(legs.map(async (currLeg, j, legsArray) => {
@@ -512,14 +558,21 @@ function parseBusRoutes(busRoutes: Array<Object>, destinationName: string): Prom
         };
       }));
 
+      const startName = directions.length > 0 ? directions[0].name : null;
+      const endName = directions.length > 0 ? directions[directions.length - 1].name : null;
+
       return {
         arrivalTime: arriveTime,
         boundingBox,
         departureTime,
         directions,
         endCoords,
+        endName,
         numberOfTransfers: busRoute.transfers,
         startCoords,
+        startName,
+        totalDuration,
+        travelDistance,
       };
     } catch (error) {
       throw new Error(
--- a/src/utils/ParseRouteUtils.js
+++ b/src/utils/ParseRouteUtils.js
@@ -8,6 +8,10 @@ import LogUtils from './LogUtils';
 import RealtimeFeedUtils from './RealtimeFeedUtils';
 import RequestUtils from './RequestUtils';
 
+const DIRECTION_TYPE = {
+  DEPART: 'depart',
+  WALK: 'walk',
+};
 const ONE_DAY_IN_MS = 86400000;
 const ONE_HOUR_IN_MS = 3600000;
 const ONE_MIN_IN_MS = 60000;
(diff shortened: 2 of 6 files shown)


src/routers/v1/MultiRouteRouter.js
```diff
@@ -35,8 +35,10 @@ class MultiRouteRouter extends ApplicationRouter<Array<Object>> {
       return [];
     }
 
+    const originName = 'Current Location';
```
young-k: why not define this up above, and shouldn't it be all caps?

src/utils/ParseRouteUtils.js
```diff
@@ -8,6 +8,10 @@ import LogUtils from './LogUtils';
 import RealtimeFeedUtils from './RealtimeFeedUtils';
 import RequestUtils from './RequestUtils';
 
+const DIRECTION_TYPE = {
```
young-k: i like this change :very-nice-kevin:

src/utils/ParseRouteUtils.js
```diff
@@ -261,6 +265,49 @@ function generateBoundingBox(path: Object): Object {
...
 
+/**
+ * Returns the distance between [startCoords] and [endCoords] in miles
+ * @param startCoords
+ * @param endCoords
+ * @returns {number}
+ */
+function getDistanceInMiles(startCoords: Object, endCoords: Object): number {
+  const lat1 = startCoords.lat;
+  const lon1 = startCoords.long;
```
young-k: prefer `startLat` and `startLong`, `lon1` is confusing

src/utils/ParseRouteUtils.js
```diff
@@ -261,6 +265,49 @@ function generateBoundingBox(path: Object): Object {
...
+  const radlat2 = Math.PI * lat2 / 180;
+  const theta = lon1 - lon2;
+  const radtheta = Math.PI * theta / 180;
+  let dist = Math.sin(radlat1) * Math.sin(radlat2) + Math.cos(radlat1) * Math.cos(radlat2) * Math.cos(radtheta);
+  if (dist > 1) {
+    dist = 1;
+  }
+  dist = Math.acos(dist);
+  dist = dist * 180 / Math.PI;
+  dist = dist * 60 * 1.1515;
```
young-k: 1.1515 needs a comment, and prefer all of these to be on one line with an explanation

src/utils/ParseRouteUtils.js
```diff
@@ -261,6 +265,49 @@ function generateBoundingBox(path: Object): Object {
...
+  return dist;
```
young-k: in general this function needs more comments

src/utils/ParseRouteUtils.js
```diff
@@ -261,6 +265,49 @@ function generateBoundingBox(path: Object): Object {
...
+    return 0;
+  }
+
+  const radlat1 = Math.PI * lat1 / 180;
+  const radlat2 = Math.PI * lat2 / 180;
+  const theta = lon1 - lon2;
+  const radtheta = Math.PI * theta / 180;
+  let dist = Math.sin(radlat1) * Math.sin(radlat2) + Math.cos(radlat1) * Math.cos(radlat2) * Math.cos(radtheta);
+  if (dist > 1) {
+    dist = 1;
```
young-k: why?

src/utils/ParseRouteUtils.js
```diff
@@ -261,6 +265,49 @@ function generateBoundingBox(path: Object): Object {
...
+ * Returns the difference between [departureTime] and [arrivalTime] in minutes.
+ * @param departureTime
+ * @param arrivalTime
+ * @returns {number}
+ */
+function getDifferenceInMinutes(departureTime: string, arrivalTime: string): number {
+  const departureDate = new Date(departureTime);
+  const arrivalDate = new Date(arrivalTime);
+  const diffInMs = arrivalDate - departureDate;
+  const diffInMins = Math.round(((diffInMs % ONE_HOUR_IN_MS) % ONE_HOUR_IN_MS) / ONE_MIN_IN_MS);
```
young-k: why do we need to take the `%` here? add comments!

src/utils/ParseRouteUtils.js
```diff
@@ -271,7 +318,13 @@ function convertMillisecondsToISOString(dateInMs: number): string {
   return `${new Date(dateInMs).toISOString().split('.')[0]}Z`;
 }
 
-function parseWalkingRoute(data: any, dateMs: number, destinationName: string, isArriveBy: boolean): Object {
+function parseWalkingRoute(
+  data: any,
```
young-k: prefer not having `any` as a type

src/utils/ParseRouteUtils.js
```diff
@@ -512,14 +587,53 @@ function parseBusRoutes(busRoutes: Array<Object>, destinationName: string): Prom
...
+          // If the direction type is walking, the stopName should be the name of the stop we just got off of.
+          stopName = previousStopName;
+        }
+        return {
+          direction,
+          stayOnBusForTransfer,
+          stopName,
+        };
+      }).filter(summary => summary !== null);
+      // The last element in routeSummary should always be an object where stopName is the destination name.
```
young-k: can we add a newline above this line just to make the separation more clear

src/utils/ParseRouteUtils.js
```diff
@@ -512,14 +587,53 @@ function parseBusRoutes(busRoutes: Array<Object>, destinationName: string): Prom
...
+      const endName = directions.length > 0 ? directions[directions.length - 1].name : null;
+
+      let previousStopName;
+      const routeSummary = directions.map((dir, index) => {
+        const { name, type, stayOnBusForTransfer } = dir;
+        // We want to ignore the intial walking direction
+        if (index === 0 && type === DIRECTION_TYPE.WALK) {
+          return null;
+        }
+        const direction = {
```
young-k: define `direction` on one line

src/utils/ParseRouteUtils.js
```diff
@@ -512,14 +587,53 @@ function parseBusRoutes(busRoutes: Array<Object>, destinationName: string): Prom
...
 
+      const startName = directions.length > 0 ? directions[0].name : null;
+      const endName = directions.length > 0 ? directions[directions.length - 1].name : null;
+
+      let previousStopName;
+      const routeSummary = directions.map((dir, index) => {
+        const { name, type, stayOnBusForTransfer } = dir;
+        // We want to ignore the intial walking direction
+        if (index === 0 && type === DIRECTION_TYPE.WALK) {
+          return null;
```
young-k: prefer putting return statement on the above line

src/utils/ParseRouteUtils.js
```diff
@@ -512,14 +587,53 @@ function parseBusRoutes(busRoutes: Array<Object>, destinationName: string): Prom
...
+        const direction = {
+          busNumber: null,
+          type,
+        };
+        let stopName = name;
+        if (type === DIRECTION_TYPE.DEPART) {
+          direction.busNumber = dir.routeNumber;
+          previousStopName = dir.stops[dir.stops.length - 1].name;
+        } else {
+          // If the direction type is walking, the stopName should be the name of the stop we just got off of.
```
young-k: prefer this comment on the previous line:

`If walking direction type, stopName should be the name of the stop we just got off`

src/utils/ParseRouteUtils.js
kevinchan159: i copied this function from the interweb so im not too sure what the numbers mean either

src/utils/ParseRouteUtils.js
```diff
@@ -512,14 +587,53 @@ function parseBusRoutes(busRoutes: Array<Object>, destinationName: string): Prom
         };
       }));
 
+      const startName = directions.length > 0 ? directions[0].name : null;
+      const endName = directions.length > 0 ? directions[directions.length - 1].name : null;
+
+      let previousStopName;
+      const routeSummary = directions.map((dir, index) => {
```
young-k: prefer `direction, index`

src/utils/ParseRouteUtils.js
young-k: this variable also seems to be named improperly, we're looping over the directions already, so isn't this not a `direction` object?

src/utils/ParseRouteUtils.js
young-k: not a good reason, and you should just link the relevant source if necessary

src/utils/ParseRouteUtils.js
kevinchan159: it is still a direction object. this represents a `RouteSummary` object where `direction` is a simplified version of our current directions

src/utils/ParseRouteUtils.js
alanna-zhou: oof

src/utils/ParseRouteUtils.js
```diff
@@ -304,17 +359,35 @@ function parseWalkingRoute(data: any, dateMs: number, destinationName: string, i
...
+      },
+    ];
+
     return adjustRouteTimesIfNecessary({
       arrivalTime,
       boundingBox,
       departureTime,
       directions: [direction],
       endCoords,
+      endName: destinationName,
```
alanna-zhou: why are these different tho

src/utils/ParseRouteUtils.js
```diff
@@ -304,17 +359,35 @@ function parseWalkingRoute(data: any, dateMs: number, destinationName: string, i
...
       numberOfTransfers: 0,
+      routeSummary,
       startCoords,
+      startName: originName,
```
alanna-zhou: same here

src/utils/ParseRouteUtils.js
kevinchan159: its what was requested by iOS but you're right, we can prob just change it

src/utils/ParseRouteUtils.js
```diff
@@ -512,14 +589,52 @@ function parseBusRoutes(busRoutes: Array<Object>, destinationName: string): Prom
...
+          stopName = previousStopName;
+        }
+        return {
+          direction: routeSummaryDirection,
+          stayOnBusForTransfer,
+          stopName,
+        };
+      }).filter(summary => summary !== null);
+
+      // The last element in RouteSummaryElement object should always be an object where stopName
```
young-k: prefer this comment be on one line, don't use words like `should`, be concise

src/utils/ParseRouteUtils.js
```diff
@@ -512,14 +589,52 @@ function parseBusRoutes(busRoutes: Array<Object>, destinationName: string): Prom
...
+        // We want to ignore the intial walking direction
+        if (index === 0 && type === DIRECTION_TYPE.WALK) return null;
+        const routeSummaryDirection = { busNumber: null, type };
+        let stopName = name;
+        if (type === DIRECTION_TYPE.DEPART) {
+          routeSummaryDirection.busNumber = direction.routeNumber;
+          previousStopName = direction.stops[direction.stops.length - 1].name;
+        } else { // If walking direction type, stopName should be the name of the stop we just got off
+          stopName = previousStopName;
+        }
```
young-k: add more whitespace, it's cluttered and tricky to read!

src/utils/ParseRouteUtils.js
kevinchan159: actually wait no its b/c we have `startCoords` and `endCoords` so its more consistent to have `startName` and `endName`

src/utils/ParseRouteUtils.js
```diff
@@ -512,14 +590,52 @@ function parseBusRoutes(busRoutes: Array<Object>, destinationName: string): Prom
...
+
+        return {
+          direction: routeSummaryDirection,
+          stayOnBusForTransfer,
+          stopName,
+        };
+      }).filter(summary => summary !== null);
```
meganle: You can do `.filter(Boolean)`

src/utils/RouteUtils.js
```diff
@@ -129,6 +129,7 @@ async function createFinalBusRoutes(
  * @returns {Promise<Object>}
  */
 async function getParsedWalkingAndBusRoutes(
+  originName: string,
```
yuna878: Update docs?

src/utils/ParseRouteUtils.js
```diff
@@ -271,7 +320,13 @@ function convertMillisecondsToISOString(dateInMs: number): string {
   return `${new Date(dateInMs).toISOString().split('.')[0]}Z`;
 }
 
-function parseWalkingRoute(data: any, dateMs: number, destinationName: string, isArriveBy: boolean): Object {
+function parseWalkingRoute(
```
yuna878: docs?

PR #235, opened 2019-05-02
Update /search endpoint to return lat and long in addition to placeID for googlePlace

alanna-zhou:
### Now a request to `/search` looks like this: 
Screen Shot 2019-05-03 at 2 36 35 PM

### DIFF
--- a/src/routers/v1/SearchRouter.js
+++ b/src/routers/v1/SearchRouter.js
@@ -5,13 +5,18 @@ import type Request from 'express';
 import AllStopUtils from '../../utils/AllStopUtils';
 import ApplicationRouter from '../../appdev/ApplicationRouter';
 import RequestUtils from '../../utils/RequestUtils';
+import Constants from '../../utils/Constants';
 
 const BUS_STOP = 'busStop';
-const cacheOptions = {
+const queryToPredictionsCacheOptions = {
   max: 10000, // Maximum size of cache
   maxAge: 1000 * 60 * 60 * 24 * 5, // Maximum age in milliseconds
 };
-const cache = LRU(cacheOptions);
+const placeIDToCoordsCacheOptions = {
+  max: 10000, // Maximum size of cache
+};
+const queryToPredictionsCache = LRU(queryToPredictionsCacheOptions);
+const placeIDToCoordsCache = LRU(placeIDToCoordsCacheOptions);
 const GOOGLE_PLACE = 'googlePlace';
 const GOOGLE_PLACE_LOCATION = '42.4440,-76.5019';
 const MIN_FUZZ_RATIO = 75;
@@ -31,7 +36,7 @@ class SearchRouter extends ApplicationRouter<Array<Object>> {
     }
 
     const query = req.body.query.toLowerCase();
-    const cachedValue = cache.get(query);
+    const cachedValue = queryToPredictionsCache.get(query);
 
     const allStops = await AllStopUtils.fetchAllStops();
     const filteredStops = allStops.filter(s => (
@@ -49,13 +54,12 @@ class SearchRouter extends ApplicationRouter<Array<Object>> {
       name: s.name,
     }));
 
-    if (cachedValue !== undefined) { // Return the list of googlePlaces and busStops
-      return cachedValue.concat(formattedStops);
-    }
+    // Return the list of googlePlaces and busStops
+    if (cachedValue) return cachedValue.concat(formattedStops);
 
     // not in cache
     const options = {
-      method: 'GET',
+      ...Constants.GET_OPTIONS,
       url: 'https://maps.googleapis.com/maps/api/place/autocomplete/json',
       qs: {
         input: query,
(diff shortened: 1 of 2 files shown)


src/routers/v1/SearchRouter.js
```diff
@@ -5,13 +5,15 @@ import type Request from 'express';
...
 import ApplicationRouter from '../../appdev/ApplicationRouter';
 import RequestUtils from '../../utils/RequestUtils';
+import Constants from '../../utils/Constants';
 
 const BUS_STOP = 'busStop';
 const cacheOptions = {
   max: 10000, // Maximum size of cache
   maxAge: 1000 * 60 * 60 * 24 * 5, // Maximum age in milliseconds
 };
 const cache = LRU(cacheOptions);
```
young-k: can we rename this cache to be more descriptive since we have two?

src/routers/v1/SearchRouter.js
```diff
@@ -5,13 +5,15 @@ import type Request from 'express';
 import AllStopUtils from '../../utils/AllStopUtils';
 import ApplicationRouter from '../../appdev/ApplicationRouter';
 import RequestUtils from '../../utils/RequestUtils';
+import Constants from '../../utils/Constants';
```
young-k: alphabetize imports

src/routers/v1/SearchRouter.js
```diff
@@ -72,22 +74,62 @@ class SearchRouter extends ApplicationRouter<Array<Object>> {
...
-        placeID: p.place_id,
+
+      const googlePredictions = await Promise.all(predictions.map(async (p): Promise<Object> => {
+        const placeIDCoords = await getCoordsForPlaceID(p.place_id);
+        return {
+          type: GOOGLE_PLACE,
+          detail: p.structured_formatting.secondary_text,
+          name: p.structured_formatting.main_text,
+          lat: placeIDCoords.lat,
+          long: placeIDCoords.long,
```
young-k: alphabetize these params

src/routers/v1/SearchRouter.js
```diff
@@ -72,22 +74,62 @@ class SearchRouter extends ApplicationRouter<Array<Object>> {
...
+        };
       }));
-      const filteredPredictions = getFilteredPredictions(googlePredictions, formattedStops);
-      cache.set(query, filteredPredictions);
 
-      // Return the list of googlePlaces and busStops
-      return filteredPredictions.concat(formattedStops);
+      if (await googlePredictions) {
```
young-k: why do we need to `await` here,  `googlePredictions` was already `await`-ed

src/routers/v1/SearchRouter.js
```diff
@@ -72,22 +74,62 @@ class SearchRouter extends ApplicationRouter<Array<Object>> {
...
+        return filteredPredictions.concat(formattedStops);
+      }
     }
     return [];
   }
 }
 
+async function getCoordsForPlaceID(placeID: String): Object {
+  const cachedValue = placeIDToCoordsCache.get(placeID);
+  if (cachedValue !== undefined) { // Return an object of lat and long
```
young-k: would the following snippet not be the same thing? (it might not, javascript handles false-y values strangely sometimes)

```
if (cachedValue) return cachedValue;
```

src/routers/v1/SearchRouter.js
```diff
@@ -72,22 +74,62 @@ class SearchRouter extends ApplicationRouter<Array<Object>> {
...
+    const placeIDDetailsResult = JSON.parse(placeIDDetailsRequest);
+    const placeIDCoords = {
+      lat: placeIDDetailsResult.result.geometry.location.lat,
+      long: placeIDDetailsResult.result.geometry.location.lng,
+    };
+
+    placeIDToCoordsCache.set(placeID, placeIDCoords);
+    return placeIDCoords;
+  }
+  return {
```
young-k: we should probably log if this request fails and we need to return `null`

src/routers/v1/SearchRouter.js
```diff
@@ -5,13 +5,15 @@ import type Request from 'express';
...
 import RequestUtils from '../../utils/RequestUtils';
+import Constants from '../../utils/Constants';
 
 const BUS_STOP = 'busStop';
 const cacheOptions = {
   max: 10000, // Maximum size of cache
   maxAge: 1000 * 60 * 60 * 24 * 5, // Maximum age in milliseconds
 };
 const cache = LRU(cacheOptions);
+const placeIDToCoordsCache = LRU(cacheOptions);
```
young-k: should this cache have a `maxAge`, or can we just store placeIDs indefinitely?

kevinchan159: the placeID to coordinate shouldn't change so i dont need we need a maxAge

src/routers/v1/SearchRouter.js
aliciaxw: +1

src/routers/v1/SearchRouter.js
aliciaxw: the snippet should be the same i think, even an empty object (if `cachedValue` could ever be that) is truthy

PR #245, opened 2019-09-02
Remove unnecessary files

young-k:
Removing:

- `travis.yml` and `docker_push`, we'll be cutting releases manually
- `graphhopper-transit-extension`, it appears that `0.11` bundles walking & routing together, so we don't need this.
- `logs/`, I'm not sure why this was committed in the first place.

Didn't test yet, but will build an image and test on the devserver prior to merging!


young-k:
Will wait to merge until `production` is stable, so I know whether or not this has caused anything to break

PR #246, opened 2019-09-02
Utilize Graphhopper 0.11 for walking directions

meganle:
- Update Dockerfile to use Graphhopper 0.11 (even though 0.11 is currently deployed)
- Sometimes the Graphhopper (0.11) bus service will return walking directions in addition to bus directions (if the distance is close enough)
- If the bus service doesn't return walking directions, then we can make a separate request to the Grasshopper walking service to get directions.
- This is good cause we don't have to make extra requests to the walking service if it's not necessary
- Also fixes bug where we get 3 walking directions by filtering walking directions from the buffered requests

Currently deployed and tested on the dev server with the beta app

### DIFF
--- a/src/utils/GraphhopperUtils.js
+++ b/src/utils/GraphhopperUtils.js
@@ -143,14 +143,13 @@ function busRoutesAreEqual(busRouteA: Object, busRouteB: Object): boolean {
 function getValidBusRoutes(busRoutes: Array<Object>, busRoutesToCheck: Array<Object>): Array<Object> {
   return busRoutesToCheck.filter((busRouteToCheck) => {
     const isDuplicateRoute = busRoutes.find(busRoute => busRoutesAreEqual(busRouteToCheck, busRoute)) !== undefined;
-    return !isDuplicateRoute;
+    return !isDuplicateRoute && busRouteToCheck.constructor.transfers !== -1;
   });
 }
 
 /**
- * Return { busRoute, walkingRoute } from graphhopper given the parameters
- * walkingRoute contains an array with length 1 containing the shortest possible walking path
- * busRoute contains an array of length 5 with possible paths
+ * Return {routes} from graphhopper given the parameters
+ * {routes} contains an array of up to length 5, with possible bus and walking routes
  * Example return object:
  {
  busRoute:
@@ -203,11 +202,11 @@ function getValidBusRoutes(busRoutes: Array<Object>, busRoutesToCheck: Array<Obj
  * @param start
  * @param departureTimeDateNow
  * @param isArriveByQuery
- * @returns {Promise<{busRoute: any, walkingRoute: any}>}
+ * @returns {Array<Object>}
  */
-async function fetchRoutes(end: string, start: string, departureTimeDateNow: string, isArriveByQuery: boolean): Object {
-  let busRoutes;
-  let walkingRoute;
+async function fetchRoutes(end: string, start: string, departureTimeDateNow: string,
+  isArriveByQuery: boolean): Array<Object> {
+  let routes;
 
   const sharedOptions = { method: 'GET', qsStringifyOptions: { arrayFormat: 'repeat' } };
   // Fetch bus routes using the current start time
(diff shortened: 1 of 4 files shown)


src/utils/GraphhopperUtils.js
```diff
@@ -143,14 +143,13 @@ function busRoutesAreEqual(busRouteA: Object, busRouteB: Object): boolean {
...
+    return !isDuplicateRoute && busRouteToCheck.constructor.transfers !== -1;
   });
 }
 
 /**
- * Return { busRoute, walkingRoute } from graphhopper given the parameters
- * walkingRoute contains an array with length 1 containing the shortest possible walking path
- * busRoute contains an array of length 5 with possible paths
+ * Return {routes} from graphhopper given the parameters
+ * routes contains an array of length 5 with possible bus and walking routes
```
young-k: `{routes} contains an array of up to length 5, with possible bus and walking routes`

src/utils/GraphhopperUtils.js
```diff
@@ -305,6 +291,33 @@ async function fetchRoutes(end: string, start: string, departureTimeDateNow: str
...
+
+/*
+* Return walkingRoute from the Graphhopper walking service given the parameters
+* walkingRoute contains the shortest possible walking path
+* @param end
+* @param start
+* @returns Object
+*/
+async function fetchWalkingRoute(end: string, start: string): Object {
+  let walkingRoute;
```
young-k: probably better to just have this be a `const` and `return` in both cases of the `if/else` instead, but up to you!

src/utils/RouteUtils.js
```diff
@@ -137,19 +137,73 @@ async function getParsedWalkingAndBusRoutes(
...
+ *
+ * @param originName
+ * @param destinationName
+ * @param end
+ * @param start
+ * @param departureTimeQuery
+ * @param isArriveBy
+ * @returns {Promise<Object>}
+ */
+async function getParsedWalkingRoute(
```
young-k: Should we have this? Does `graphhopper`, with this version, ever not return walking routes?

meganle: Yeah if the distance is far (prob > 1 mile) it just won't return any walking directions

src/utils/GraphhopperUtils.js
```diff
@@ -143,14 +143,13 @@ function busRoutesAreEqual(busRouteA: Object, busRouteB: Object): boolean {
 function getValidBusRoutes(busRoutes: Array<Object>, busRoutesToCheck: Array<Object>): Array<Object> {
   return busRoutesToCheck.filter((busRouteToCheck) => {
     const isDuplicateRoute = busRoutes.find(busRoute => busRoutesAreEqual(busRouteToCheck, busRoute)) !== undefined;
-    return !isDuplicateRoute;
+    return !isDuplicateRoute && busRouteToCheck.constructor.transfers !== -1;
   });
```
alanna-zhou: this isn't part of the PR but why would there be duplicate routes?

src/utils/GraphhopperUtils.js
```diff
@@ -285,7 +271,7 @@ async function fetchRoutes(end: string, start: string, departureTimeDateNow: str
 
   if (busRouteBufferedFirstRequest && busRouteBufferedFirstRequest.statusCode < 300) {
     const bufferedBusRoutes = JSON.parse(busRouteBufferedFirstRequest.body).paths;
-    busRoutes = getValidBusRoutes(busRoutes, bufferedBusRoutes).concat(busRoutes);
+    routes = getValidBusRoutes(routes, bufferedBusRoutes).concat(routes);
```
alanna-zhou: also not in this pr but is there a reason why this can't be `routes = routes.concat(getValidBusRoutes(routes, bufferedBusRoutes)`?

src/utils/ParseRouteUtils.js
```diff
@@ -660,6 +660,6 @@ function parseBusRoutes(busRoutes: Array<Object>, originName: string, destinatio
 
 export default {
   condenseRoute,
-  parseBusRoutes,
+  parseRoutes,
```
alanna-zhou: wait so just so im understanding this correctly, parseBusRoutes got renamed to parseRoutes bc it could potentially have walking routes, but you still have the parseWalkingRoute in the case that ghopper doesn't give a walking route for a short distance?

src/utils/GraphhopperUtils.js
meganle: since we do buffered requests, a request for 2 minutes after might return the same route as making a request for now

src/utils/GraphhopperUtils.js
meganle: the buffered requests return routes before the start time, so I'm guessing we want those routes first

src/utils/ParseRouteUtils.js
meganle: yep it's not guaranteed that ghopper will return both walking and bus routes

Issue #177, opened 2019-02-21
Get rid of traditional for loops


young-k:
Probably a few in this codebase that we should change to `forEach` as @cswenberg noted in my PR, just `git grep` it and make sure we get them all!

PR #254, opened 2019-09-13
Replace last remove-able for loop with forEach

alanna-zhou:
#177 I've addressed the last traditional for loop that could be replaced with a forEach

Sorry for the duplicate PR but I'll be breaking up my first PR into chunks because the massive change is breaking the dev server and I don't know why, so I'm testing each chunk. I've tested this change on the dev server (transit-dev:alanna1047), and I know for sure that it works.

Once I get all my other PRs in, I'll delete my old one.

When doing a search in VSCode, there are two more traditional `for` loops left because the first one we modify the array based on the index (so it didn't make sense to me to change that to a `forEach`) and the second one we're not looping through an array at all:
### Two results from searching
Screen Shot 2019-09-14 at 4 10 47 PM

### uses index to modify array while iterating through it
Screen Shot 2019-09-14 at 4 11 15 PM

### just loops through numbers, not an array
Screen Shot 2019-09-14 at 4 11 34 PM

### DIFF
--- a/src/appdev/ApplicationAPI.js
+++ b/src/appdev/ApplicationAPI.js
@@ -39,9 +39,9 @@ class ApplicationAPI {
     AppDevUtilities.tryCheckAppDevURL(this.getPath());
     const middleware = this.middleware();
     const routerGroups = this.routerGroups();
-    for (let i = 0; i < middleware.length; i++) {
-      this.express.use(middleware[i]);
-    }
+    middleware.forEach((m) => {
+      this.express.use(m);
+    });
 
     Object.keys(routerGroups).forEach((version) => {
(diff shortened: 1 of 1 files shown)


young-k:
can you also screenshot a `find` or a `grep` showing that there aren't any results for `for ...` in the codebase?

young-k:
Can't we get rid of the second one? Maybe:

```
for (const x of Array(retryCount).keys()) {
  ...
}
```

kq-li:
@young-k initializing an array with that length and iterating its indices isn't really more clear and definitely isn't more performant -- imo this is a valid use case for a regular for loop

PR #255, opened 2019-09-13
Clean up constants

alanna-zhou:
Centralized all the constants in one file, and removed the postman token because it was useless.

Again, sorry for the duplicate PR--for some reason, using the RETRY_COUNT constant in RequestUtils.js broke everything. Will be deleting the original PR once this gets merged in.

Currently testing on dev server (transit-dev:alanna) because it sometimes works and sometimes doesn't work

### DIFF
--- a/src/utils/AllStopUtils.js
+++ b/src/utils/AllStopUtils.js
@@ -3,6 +3,9 @@ import Constants from './Constants';
 import { PYTHON_APP } from './EnvUtils';
 import RequestUtils from './RequestUtils';
 
+// TODO: investigate why importing from Constants.js results in failure of getting routes
+const DEG_EQ_PRECISION = 5; // 5 degrees of precision is about a 1.1 meters, is a stop
+
 /**
  * Used for finding stops at or nearby a point
  * Use getPrecisionMap(precision) to access.
@@ -62,7 +65,7 @@ async function fetchAllStops() {
 
 function fetchPrecisionMaps() {
   const maps = {};
-  maps[Constants.DEG_EQ_PRECISION] = getPrecisionMap(Constants.DEG_EQ_PRECISION);
+  maps[DEG_EQ_PRECISION] = getPrecisionMap(DEG_EQ_PRECISION);
   return maps;
 }
 
@@ -72,8 +75,9 @@ function fetchPrecisionMaps() {
  * @param degreesPrecision
  * @returns {Promise<void>}
  */
-async function getPrecisionMap(degreesPrecision: ?number = Constants.DEG_EQ_PRECISION) {
-  if (degreesPrecision < Constants.DEG_MIN_PRECISION
+async function getPrecisionMap(degreesPrecision: ?number = DEG_EQ_PRECISION) {
+  if (
+    degreesPrecision < Constants.DEG_MIN_PRECISION
     || degreesPrecision > Constants.DEG_MAX_PRECISION
   ) {
     return null;
@@ -125,7 +129,7 @@ async function isStopsWithinPrecision(point: Object, degreesPrecision: ?number =
  * @returns {Promise<boolean>}
  */
 function isStop(point: Object) {
-  return isStopsWithinPrecision(point, Constants.DEG_EQ_PRECISION);
+  return isStopsWithinPrecision(point, DEG_EQ_PRECISION);
 }
 
 export default {
--- a/src/utils/AllStopUtils.js
+++ b/src/utils/AllStopUtils.js
@@ -3,15 +3,6 @@ import Constants from './Constants';
 import { PYTHON_APP } from './EnvUtils';
 import RequestUtils from './RequestUtils';
 
-const SEC_IN_MS = 1000;
-const MIN_IN_MS = SEC_IN_MS * 60;
-const HOUR_IN_MS = MIN_IN_MS * 60;
-const DEG_EXACT_PRECISION = 6; // 6 degrees of precision is about a 111 mm, is exact point
-const DEG_EQ_PRECISION = 5; // 5 degrees of precision is about a 1.1 meters, is a stop
-const DEG_NEARBY_PRECISION = 4; // 4 degrees of precision is about 11 meters, stop nearby
-const DEG_WALK_PRECISION = 3; // 3 degrees of precision is about 111 meters, stop walkable
-const DEG_KM_PRECISION = 2; // 3 degrees of precision is about 1 km, stop barely walkable
-
 /**
  * Used for finding stops at or nearby a point
  * Use getPrecisionMap(precision) to access.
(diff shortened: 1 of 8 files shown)


src/utils/Constants.js
```diff
@@ -1,11 +1,91 @@
...
+const WALK_SPEED = 3.0;
+
+/* Degrees Precision constants */
+const DEG_EQ_PRECISION = 5; // 5 degrees of precision is about a 1.1 meters, is a stop
+const DEG_EXACT_PRECISION = 6; // 6 degrees of precision is about a 111 mm, is exact point
+const DEG_KM_PRECISION = 2; // 3 degrees of precision is about 1 km, stop barely walkable
+const DEG_MAX_PRECISION = 6;
+const DEG_MIN_PRECISION = 1;
+const DEG_NEARBY_PRECISION = 4; // 4 degrees of precision is about 11 meters, stop nearby
+const DEG_WALK_PRECISION = 3; // 3 degrees of precision is about 111 meters, stop walkable
```
young-k: I think we should actually not alphabetize and have these in order, it'd be clearer for the future

src/utils/Constants.js
```diff
@@ -1,11 +1,91 @@
...
+/* Delay Buffer constants */
+// buffer to account for routes in past 20 minutes with delays
+const FIRST_DELAY_BUFFER_IN_MINUTES = 20;
+// additional buffer to account for time needed to walk from current location to bus stop
+const SECOND_DELAY_BUFFER_IN_MINUTES = 40;
+
+/* Distance & Speed constants */
+// > 3.0 suggests getting off bus earlier and walk half a mile instead of waiting longer
+const MAX_WALK_DIST_PER_LEG = 2000;
+const AUTOCOMPLETE_RADIUS = 24140;
```
young-k: we probably need a comment here, if you're not sure could you add a TODO here?

src/utils/Constants.js
```diff
@@ -1,11 +1,91 @@
...
+  max: 1000, // max 1000 routes in storage
+  maxAge: 1000 * 60 * 15, // max age 15 minutes
+};
+const QUERY_PREDICTIONS_CACHE_OPTIONS = {
+  max: 10000, // Maximum size of cache
+  maxAge: 1000 * 60 * 60 * 24 * 5, // Maximum age in milliseconds
+};
+
+/* Count constants */
+const MIN_FUZZ_RATIO = 75;
```
young-k: what is a fuzz ratio? probably useful to put that here

src/utils/Constants.js
```diff
@@ -1,11 +1,91 @@
...
+  max: 10000,
+  maxAge: 1000 * 60 * 60 * 24 * 5,
+};
+const ROUTES_CALC_CACHE_OPTIONS = {
+  max: 1000, // max 1000 routes in storage
+  maxAge: 1000 * 60 * 15, // max age 15 minutes
+};
+const QUERY_PREDICTIONS_CACHE_OPTIONS = {
+  max: 10000, // Maximum size of cache
+  maxAge: 1000 * 60 * 60 * 24 * 5, // Maximum age in milliseconds
```
young-k: for the middle `maxAge` we put a comment with the number of minutes, can we do something similar for the first one and this one as well?

src/utils/AllStopUtils.js
```diff
@@ -81,8 +72,9 @@ function fetchPrecisionMaps() {
  * @param degreesPrecision
  * @returns {Promise<void>}
  */
-async function getPrecisionMap(degreesPrecision: ?number = DEG_EQ_PRECISION) {
-  if (degreesPrecision < 1 || degreesPrecision > 6) {
+async function getPrecisionMap(degreesPrecision: ?number = Constants.DEG_EQ_PRECISION) {
+  if (degreesPrecision < Constants.DEG_MIN_PRECISION
+    || degreesPrecision > Constants.DEG_MAX_PRECISION) {
```
kq-li:
```
if (
  degreesPrecision < Constants.DEG_MIN_PRECISION
  || degreesPrecision > Constants.DEG_MAX_PRECISION
) {
```

src/utils/Constants.js
kq-li: these don't really seem to reference the same value, how would you order them?

young-k: oh i meant in terms of like 1-6 in terms of value, but I don't have a strong preference!

kq-li: oh duh LOL me neither

src/utils/Constants.js
```diff
@@ -24,17 +25,18 @@ const SECOND_DELAY_BUFFER_IN_MINUTES = 40;
 /* Distance & Speed constants */
 // > 3.0 suggests getting off bus earlier and walk half a mile instead of waiting longer
 const MAX_WALK_DIST_PER_LEG = 2000;
+// The distance (in meters) within which to return Google place results for autocomplete.
 const AUTOCOMPLETE_RADIUS = 24140;
 const WALK_SPEED = 3.0;
 
 /* Degrees Precision constants */
+const DEG_MIN_PRECISION = 1;
+const DEG_KM_PRECISION = 2; // 3 degrees of precision is about 1 km, stop barely walkable
```
kq-li: should this comment say 2 degrees?

src/utils/Constants.js
```diff
@@ -24,17 +25,18 @@ const SECOND_DELAY_BUFFER_IN_MINUTES = 40;
...
+const DEG_WALK_PRECISION = 3; // 3 degrees of precision is about 111 meters, stop walkable
+const DEG_NEARBY_PRECISION = 4; // 4 degrees of precision is about 11 meters, stop nearby
 const DEG_EQ_PRECISION = 5; // 5 degrees of precision is about a 1.1 meters, is a stop
```
kq-li: `about a 1.1` => `about 1.1`

src/utils/Constants.js
```diff
@@ -24,17 +25,18 @@ const SECOND_DELAY_BUFFER_IN_MINUTES = 40;
...
 const DEG_EXACT_PRECISION = 6; // 6 degrees of precision is about a 111 mm, is exact point
```
kq-li: `about a 111` => `about 111`

src/utils/AllStopUtils.js
```diff
@@ -74,7 +74,8 @@ function fetchPrecisionMaps() {
  */
 async function getPrecisionMap(degreesPrecision: ?number = Constants.DEG_EQ_PRECISION) {
   if (degreesPrecision < Constants.DEG_MIN_PRECISION
```
kq-li: newline after the initial `if (` is cleaner imo but your call

young-k: I agree with @kq-li here!

src/utils/Constants.js
```diff
@@ -1,12 +1,17 @@
...
   max: 10000, // Max 10000 autocomplete results
   maxAge: 1000 * 60 * 60 * 24 * 5, // Max age in 5 days
 };
+
+// A cache containing saved routes IDs to routes
 const ROUTES_CALC_CACHE_OPTIONS = {
   max: 1000, // Max 1000 routes
   maxAge: 1000 * 60 * 15, // Max age in 15 minutes
 };
+// A cache mapping a /search's request query string to its Google autocomplete predictions
```
kq-li: this newline spacing is inconsistent with the above change

PR #257, opened 2019-09-20
Return routes that end at a time before the arriveBy time

### LINKED ISSUE #253: Arrive by not showing routes that should show
Example: Route exists at 6:52 that arrives at 7:01 (Image A). Therefore, setting arrive by 7:05 should display this route. However, it isn't displayed (Image B).

Image A:
Screen Shot 2019-09-11 at 6 05 26 PM

Image B:
Screen Shot 2019-09-11 at 6 05 42 PM

Proof request is made with arriveBy parameter:
Screen Shot 2019-09-11 at 6 04 16 PM

Screen Shot 2019-09-11 at 6 09 33 PM

alanna-zhou:
Fixed #253, which was that when a client indeed passes in a boolean `arriveBy`, `true`, when that gets passed in as a request to the backend, the type of `arriveBy` is now tragically a String, `'true'`. The unfortunate disaster is we only check for `'1'` or `true`, but not `'true'`.

Tested on dev server and ensured that we do return routes that have an ending time before the `arriveBy` time.

# I tested this on the dev server (transit-dev:alanna-arrive) and it works, and here is the evidence:

Here is the GET request I used: https://transit-testflight.cornellappdev.com/api/v1/route?arriveBy=true&end=42.4426153,-76.50927779999999&start=42.4439852,-76.48287359999999&time=1568948285&destinationName=Dufield%20Hall&originName=Statler%20Hall

Notice the arriveBy is true and the time is 1568948285, which is 
Screen Shot 2019-09-19 at 11 06 37 PM

# Here was what we were returning before:
Screen Shot 2019-09-19 at 11 06 11 PM

# And this is what is returned now:
Screen Shot 2019-09-19 at 11 05 35 PM

# However, you may notice that the UI doesn't reflect that. 
I suspect that iOS isn't actually passing in the true parameter, because I'm not sure why I would get my results by making a direct request to backend vs testing it on the Transit Beta App. @Omarrasheed @usubzero

### DIFF
--- a/src/routers/v1/RouteRouter.js
+++ b/src/routers/v1/RouteRouter.js
@@ -27,7 +27,7 @@ class RouteRouter extends ApplicationRouter<Array<Object>> {
       uid,
     } = params;
 
-    const isArriveBy = (arriveBy === '1' || arriveBy === true);
+    const isArriveBy = (arriveBy === '1' || arriveBy === true || arriveBy === 'true' || arriveBy === 'True');
     const routes = await RouteUtils.getRoutes(originName, destinationName, end, start, time, isArriveBy);
     const containsTransfer = routes.find(route => RouteUtils.routeContainsTransfer(route)) !== undefined;
     if (routes.length > 0 && containsTransfer) {


kevinchan159:
@lucyxubroad @usubzero @Omarrasheed can we add an issue for this if we don't have one already? we shouldn't have to have these changes

PR #260, opened 2019-09-23
Add ApplePlaces & AppleSearchRouter

kevinchan159:
To convert from using GooglePlaces API to using Apple's FREE MapKit, here is my current plan:

When a user makes a search on iOS for a query, i.e. `Duff`, client will hit `/appleSearch` which will return an array of `busStops` AND an array of `applePlaces` if those places exist in our cache. If it doesn't, we just return an array of `busStops`. On client-side, when we see that server did not return any `applePlaces`, we will do the lookup for ApplePlaces and then send that array of `applePlaces` to `/applePlaces` which will just store these places in the cache. Now, the next time a user types in `Duff`, server will just return both the `busStops` and `applePlaces` and client will not need to do the lookup for ApplePlaces.

This PR accomplishes this plan by creating an `AppleSearchRouter` and `ApplePlacesRouter`. I'm honestly not sure what the best names for these two were so suggestions are appreciated!

### DIFF
--- a/src/utils/SearchUtils.js
+++ b/src/utils/SearchUtils.js
@@ -1,12 +1,21 @@
 // @flow
+import fuzz from 'fuzzball';
 import LRU from 'lru-cache';
+import AllStopUtils from './AllStopUtils';
 import RequestUtils from './RequestUtils';
 import Constants from './Constants';
 
+const BUS_STOP = 'busStop';
+const MIN_FUZZ_RATIO = 75;
 const placeIDToCoordsCacheOptions = {
   max: 10000, // Maximum size of cache
 };
 const placeIDToCoordsCache = LRU(placeIDToCoordsCacheOptions);
+const queryToPredictionsCacheOptions = {
+  max: 10000, // Maximum size of cache
+  maxAge: 1000 * 60 * 60 * 24 * 5, // Maximum age in milliseconds
+};
+const queryToPredictionsCache = LRU(queryToPredictionsCacheOptions);
 
 async function getCoordsForPlaceID(placeID: String): Object {
   const cachedValue = placeIDToCoordsCache.get(placeID);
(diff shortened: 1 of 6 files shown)


src/utils/SearchUtils.js
```diff
@@ -1,12 +1,21 @@
...
 import RequestUtils from './RequestUtils';
 import Constants from './Constants';
 
+const BUS_STOP = 'busStop';
+const MIN_FUZZ_RATIO = 75;
 const placeIDToCoordsCacheOptions = {
   max: 10000, // Maximum size of cache
 };
 const placeIDToCoordsCache = LRU(placeIDToCoordsCacheOptions);
+const queryToPredictionsCacheOptions = {
```
alanna-zhou: why don't you just combine the two caches into one

kevinchan159: the current cache in `/search` stores Google specific info like `placeID` and we have to keep those for users who don't update to this new version when we push it out.

kevinchan159:
> has this been tested on the dev server yet?

not yet but gonna test as I develop on iOS. will test before merging tho!

kevinchan159:
> wait also where are you making the request to the mapkit?

Request to MapKit is done on iOS side

src/utils/SearchUtils.js
alanna-zhou: i understand, but what i meant is couldn't you just have one cache that holds lots of info per query including placeid?

alanna-zhou:
> > has this been tested on the dev server yet?
> 
> not yet but gonna test as I develop on iOS. will test before merging tho!

gotcha! let me know if you are, because i'm currently making my latest changes sit on the dev server for a bit to make sure that it's save to be on production :)

src/utils/SearchUtils.js
kevinchan159: oh i understand what you mean. lets say we have a query `Duff` and and the user who searches it is using Apple MapKit. Then, inside the cache, we would have `Duff: [{name, lat, long}...]` w/o `placeID`'s. Lets say the next user then queries `Duff` but has not updated and is using `GoogleMaps`, if we were using the same cache we would have to check if the values for `Duff` include `placeID`. If they don't, as in this case, then we would have to do the GoogleMaps lookup and overwrite the cache. My thinking was just that since the ApplePlaces and GooglePlaces objects contain different keys, having it as two separate caches made sense. lmk if that sounds reasonable or not

PR #265, opened 2019-09-29
Update README to better explain how to run locally

esfinkel:
Update README to better explain how to run locally, using additional information provided by @alanna-zhou .

Addresses the [issue raised here](https://github.com/cuappdev/ithaca-transit-backend/issues/259)

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,10 +10,10 @@ Download on the App Store [here](https://itunes.apple.com/app/id1290883721)!
 Please ensure `npm` ([npm](https://www.npmjs.com/get-npm)) and `docker` ([docker](https://www.docker.com/)) are installed by checking `npm -v` and `docker -v` and that Docker is running.
 
 ## Environment Variables
-It's recommended to use [`direnv`](https://direnv.net):
+It's recommended to use [`direnv`](https://direnv.net).
 The required environment variables for this API are the following:
 
-To use `direnv` with this repository, run the following and set the variables appropriately.
+To use `direnv` with this repository, run the following and then set the variables appropriately.
 
 ```bash
 cp envrc.template .envrc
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,14 +28,20 @@ Environment variable values can be found by asking a member of Cornell AppDev.
...
 npm run build:dev
 npm run start:microservices
 npm run start:node
 ```
 
+```build:dev``` builds node, ```start:microservices``` starts up its microservices, and ```start:node``` runs node. 
+
+You only need to run ```npm run build:dev``` and ```npm run start:microservices``` once - the "volumes" part of docker-compose.yml allows for hot reloading. However, if you are building an image for production, make sure to do ```npm run build:dev``` right before doing that. And you will always need to run ```npm run start:node``` to run node locally.
+
+> It is helpful to understand the scripts in package.json. When you run ```npm run start:node```, it goes to find start:node's value in package.json's scripts. You can see that it's ```docker-compose up node```. This makes sense because in docker-compose.yml, you can see that the target is development. The command for starting development is ```npm run start:dev```; in package.json, its value tells us to actually execute some magic (./node_modules/.bin/webpack…). The reason why we do this is because ```npm run start:node``` lets us run on our local computers while ```npm run start:dev``` runs inside the image.
```
meganle: Maybe we can make this section more succinct! You could do something like
```suggestion
> When you run ```npm run start:node```, it actually runs ```docker-compose up node```.  On the other hand, the command for starting development is ```npm run start:dev```; in package.json, its value tells us to actually execute some magic (./node_modules/.bin/webpack…). The main difference between these two is that ```npm run start:node``` lets us run node inside the image while ```npm run start:dev``` runs on our local computer.
```
@alanna-zhou can you read this again

esfinkel: That looks better. We might still want to reference package.json in the first line, depending on whether the prospective viewer will already know to look there.

meganle: Cool sounds like a good idea!

esfinkel:
@meganle @alanna-zhou 
Reordered, abbreviated slightly, and added bullet points. I'm not super familiar with README conventions on GitHub so I appreciate your insight.

Issue #264, opened 2019-09-28
Consistent fields for all objects


Omarrasheed:
Some responses from backend, specifically from /route, tend to have some missing keys. For example, keys such as `routeNumber`, `stayOnBusForTransfer`, and `delay` don't always exist and end up throwing errors on client side.

On iOS, we're currently running some do catch blocks in case some keys are missing but ideally backend responses would be consistent, even if they're null values.
<img width="654" alt="Screen Shot 2019-09-28 at 4 42 08 PM" src="https://user-images.githubusercontent.com/36868927/65822262-00f5b400-e20f-11e9-8af1-5e66fbbdddb0.png">

PR #266, opened 2019-09-29
Add delay and stayOnBusForTransfer fields for complete v2/route responses

alanna-zhou:
# The issue
#264 Right now, we don't always return `delay` or `stayOnBusForTransfer` in our `directions` list that's returned by `v2\route`, which causes issues for iOS and is the reason for why the app shows a `could not fetch routes` error for only certain stops:

## Before
image

## After
image

# The fix
Thanks to @kevinchan159, I found that `mergeDirections` in `parseRouteUtils.js` was missing some fields (`delay` and `stayOnBusForTransfer`!) when it was merging routes, so I added the fields in with logic that:
1.  the combined delay is only the first bus route's delay, because that's how much the combined route is initially delayed by and is what the user needs to know in order to board the route on time
2. the combined `stayOnBusForTransfer` is only true if both routes' `stayOnBusForTransfer`s are true -- if you have to get off the bus at any point in time, the whole route's `stayOnBusForTransfer` is false! I could also get feedback on this since I only concluded this logically

# Validating the fix
Here's the before and after of our backend responses:
Screen Shot 2019-09-29 at 7 31 22 PM
Screen Shot 2019-09-29 at 7 31 29 PM

I've also tested this on `transit-dev` (`transit-dev:missing-keys`) with `Transit Beta 1.6.2 (6)` and I ran `integration` locally and got 21/21 tests pass :) 

# Another update
I also updated the `docker-compose.yml` because the new tags are required in order to run this PR locally.

### DIFF
--- a/src/utils/ParseRouteUtils.js
+++ b/src/utils/ParseRouteUtils.js
@@ -1,6 +1,7 @@
 // @flow
 import createGpx from 'gps-to-gpx';
 
+import { isNullOrUndefined } from 'util';
 import { MAP_MATCHING } from './EnvUtils';
 import AllStopUtils from './AllStopUtils';
 import GTFSUtils from './GTFSUtils';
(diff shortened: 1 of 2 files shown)


esfinkel:
1. If you're transferring from one bus to another, should the delay be "how late you are relative to what you expected", i.e. just the second bus's delay? So if both buses are 5 mins late, overall delay is 5 mins? I don't think there's a **wrong** answer, it just depends on the convention the user expects.

(Incidentally, I imagine we have it configured to not suggest routes where the first bus's delay would make you miss the second bus?)

2. That makes sense.

alanna-zhou:
> as far as your implementation logic:
> 
> 1. If you're transferring from one bus to another, should the delay be "how late you are relative to what you expected", i.e. just the second bus's delay? So if both buses are 5 mins late, overall delay is 5 mins? I don't think there's a **wrong** answer, it just depends on the convention the user expects.
> 
> (Incidentally, I imagine we have it configured to not suggest routes where the first bus's delay would make you miss the second bus?)
> 
> 1. The logic for stayOnBusForTransfer makes sense.

good point! I think it makes more sense to take the first bus' delay, because the user is more interested in how much their first bus is delayed by. Once they take the first bus, then they'll be concerned about the second bus' delay.

alanna-zhou:
> Wow, what a well-written PR! Is the delay what is used when it displays like `Late - 6:27 PM` or where is it used on the iOS side? Because if the directions involve taking two buses we want to show just the delay of the first bus for the boarding time vs original boarding time + total delay

Thanks @meganle! First bus delay does make more sense. Updating now~

src/utils/ParseRouteUtils.js
```diff
@@ -65,7 +66,25 @@ function mergeDirections(first, second): Object {
   const stops = first.stops.concat(second.stops);
   const tripIDs = first.tripIdentifiers.concat(second.tripIdentifiers);
 
+  // The combined delay is only the first bus route's delay, because that is
+  // how much the combined route is initially delayed by and is what the user
+  // needs to know in order to board the route on time.
+  const delay = isNullOrUndefined(first.delay) ? 0 : first.delay;
```
meganle: Should the delay be 0 or nil if `first.delay` is undefined? Does 0 signify on time?

alanna-zhou: oh wait that's a good point thanks

meganle:
@esfinkel In javascript, `{distance}` is equivalent to `{"distance": distance}` or `{distance: distance`}

Issue #262, opened 2019-09-26
Add new /delays endpoint 


alanna-zhou:
currently, iOS gets a delay for each route, which is tedious -- to fix this, iOS is planning to send an array of objects that have tripID and stopID attributes, so the backend should give a set of delays for a set of those objects

PR #268, opened 2019-10-05
Add new /delays endpoint for easier iOS handling of live tracking

alanna-zhou:
#262 

Right now iOS has to make a request for each route's delay, so now a DelaysRouter exists such that it can take in multiple tripID and stopIDs to get multiple delays:

Screen Shot 2019-10-05 at 1 02 28 AM

### DIFF
--- a/src/routers/v2/DelaysRouter.js
+++ b/src/routers/v2/DelaysRouter.js
@@ -1,0 +1,30 @@
+// @flow
+import ApplicationRouter from '../../appdev/ApplicationRouter';
+import RealtimeFeedUtils from '../../utils/RealtimeFeedUtils';
+
+class DelaysRouter extends ApplicationRouter<Object> {
+  constructor() {
+    super(['POST']);
+  }
+
+  getPath(): string {
+    return '/delays/';
+  }
+
+  async content(req): Promise<any> {
+    const rtf = await RealtimeFeedUtils.fetchRTF();
+    const delays = req.body.data.map(async ({ stopID, tripID }) => {
+      const res = await RealtimeFeedUtils.getDelayInformation(
+        stopID, tripID, rtf,
+      );
+      return {
+        stopID,
+        tripID,
+        delay: res ? res.delay : null,
+      };
+    });
+    return Promise.all(delays);
+  }
+}
+
+export default new DelaysRouter().router;
--- a/src/API.js
+++ b/src/API.js
@@ -20,6 +20,7 @@ class API extends ApplicationAPI {
       Routers.ApplePlacesRouter,
       Routers.AppleSearchRouter,
       Routers.DelayRouter,
+      Routers.DelaysRouter,
       Routers.HelloWorldRouter,
       Routers.MultiRouteRouter,
       Routers.PlacesAutocompleteRouter,
--- a/src/routers/index.js
+++ b/src/routers/index.js
@@ -3,6 +3,7 @@ import AllStopsRouter from './v1/AllStopsRouter';
 import ApplePlacesRouter from './v1/ApplePlacesRouter';
 import AppleSearchRouter from './v1/AppleSearchRouter';
 import DelayRouter from './v1/DelayRouter';
+import DelaysRouter from './v2/DelaysRouter';
 import DocsRouter from './DocsRouter';
 import HelloWorldRouter from './v1/HelloWorldRouter';
 import MultiRouteRouter from './v1/MultiRouteRouter';
@@ -20,6 +21,7 @@ export default {
   ApplePlacesRouter,
   AppleSearchRouter,
   DelayRouter,
+  DelaysRouter,
   DocsRouter,
   HelloWorldRouter,
   MultiRouteRouter,


src/routers/v2/DelaysRouter.js
```diff
@@ -0,0 +1,32 @@
...
+    const rtf = await RealtimeFeedUtils.fetchRTF();
+    const delaysPromises = delays.map(async (stopAndTripID) => {
+      const tripID = await stopAndTripID.tripID;
+      const stopID = await stopAndTripID.stopID;
+      const res = await RealtimeFeedUtils.getDelayInformation(stopID, tripID, rtf);
+      if (res) {
+        stopAndTripID.delay = res.delay;
+      } else {
+        stopAndTripID.delay = null;
+      }
```
Omarrasheed: nit: `stopAndTripID.delay = res ? res.delay : null`

young-k: debatable whether or not you accept this :joy:, I think the way you have it is more readable!

src/routers/v2/DelaysRouter.js
```diff
@@ -0,0 +1,32 @@
...
+  getPath(): string {
+    return '/delays/';
+  }
+
+  async content(req): Promise<any> {
+    const delays = req.body.data;
+    const rtf = await RealtimeFeedUtils.fetchRTF();
+    const delaysPromises = delays.map(async (stopAndTripID) => {
+      const tripID = await stopAndTripID.tripID;
+      const stopID = await stopAndTripID.stopID;
```
young-k: why do these need to be `await`ed?

src/routers/v2/DelaysRouter.js
```diff
@@ -0,0 +1,32 @@
...
+      const res = await RealtimeFeedUtils.getDelayInformation(stopID, tripID, rtf);
+      if (res) {
+        stopAndTripID.delay = res.delay;
+      } else {
+        stopAndTripID.delay = null;
+      }
+    });
+    await Promise.all(delaysPromises);
+    return delays;
```
young-k: This is confusing! `delays` is a `const`, which implicitly means that it shouldn't be modified, so someone quickly reading this code would assume that we're incorrectly returning a value that hasn't been modified since we set it.

Why not `let` assign `delays` above, and not create a new variable for `delaysPromises`? You can also return `Promise.all(delaysPromises)`, since this function itself is also `async`.

src/routers/v2/DelaysRouter.js
alanna-zhou: I actually added the awaits because both the linter and my terminal were giving me errors:
Screen Shot 2019-10-06 at 1 55 24 PM
Screen Shot 2019-10-06 at 1 55 18 PM

src/routers/v2/DelaysRouter.js
alanna-zhou: the linter actually prefers const because i'm technically not modifying it?
Screen Shot 2019-10-06 at 1 55 18 PM

src/routers/v2/DelaysRouter.js
alanna-zhou: i got rid of the variables instead and are just directly referencing the attributes now

src/routers/v2/DelaysRouter.js
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  getPath(): string {
+    return '/delays/';
+  }
+
+  async content(req): Promise<any> {
+    const delays = req.body.data;
+    const rtf = await RealtimeFeedUtils.fetchRTF();
+    delays.map(async (stopAndTripID) => {
```
kq-li: i think `stopAndTripID` could be either be renamed something else more indicative of the actual function -- `tripData` or something along those lines -- or just directly destructured -- `async ({stopID, tripID}) => {`.

src/routers/v2/DelaysRouter.js
```diff
@@ -0,0 +1,27 @@
...
+      const res = await RealtimeFeedUtils.getDelayInformation(
+        stopAndTripID.stopID, stopAndTripID.tripID, rtf,
+      );
+      stopAndTripID.delay = res ? res.delay : null;
+    });
+    return Promise.all(delays);
```
kq-li: something seems wrong here -- you throw away the returned `Promise<void>`s from this map. this might still work (maybe the mutation makes it fine?) but i feel like this should really just be immutable and read something like 
```
return {
  ...stopAndTripID,
  delay: res ? res.delay : null,
}
```
in the map callback and then you'd return the `Promise.all` of the map result rather than the original `delays`.

src/routers/v2/DelaysRouter.js
meganle: Here's a helpful article about destructuring https://wesbos.com/destructuring-objects/
It's something that is commonly used in JS!

src/routers/v2/DelaysRouter.js
alanna-zhou: i'm confused - what's wrong with modifying the array?

src/routers/v2/DelaysRouter.js
young-k: Is this error because of the `const`? Isn't it because you're not destructuring, as @meganle noted?

src/routers/v2/DelaysRouter.js
kq-li: so my main concern is that `delays` is not a promise or array of promises, making your `Promise.all` call invalid, and also that the return value of the `map` is just not being used. i think the typing issues aren't being properly caught here -- it's an example of a type-unsound but functional program

young-k: ahhh yes @kq-li is right, you should return the `map` value here, I thought it was in-place but `map` returns a new array!

Issue #14, opened 2017-09-21
[Testing] Instantiate footpath directly for each test case


Jma353:
This way we have direct control over the route + the optimalities it should represent

PR #271, opened 2019-10-27
Add route to retrieve GTFS feed info from microservice

sahamanish99:
## Overview

Beginning to tackle integration issue: https://github.com/cuappdev/integration/issues/14
However, to retrieve the GTFS feed info, we have to create an endpoint on `ithaca-transit-backend`, since `/gtfs-feed-info/` isn't public. 

## Changes Made

I added a new endpoint `/GTFSFeedInfo/` to retrieve the data from the microservice. This can be found in FeedInfoRouter. 

## Test Coverage

I tested on Postman to make sure we were getting the proper response. Upon hitting `/api/v1/GTFSFeedInfo` I get the proper response extracted from the microservice endpoint.

```
{
    "success": true,
    "data": {
        "feed_end_date": "20200118",
        "feed_start_date": "20191027",
        "feed_version": "20191027"
    }
}
```
## Related PRs or Issues

https://github.com/cuappdev/integration/issues/14


src/utils/GTFSUtils.js
```diff
@@ -12,6 +12,17 @@ async function fetchRoutes(): Object {
   return JSON.parse(data);
 }
 
+async function getFeedInfo(): Object {
+  const options = {
+    method: 'GET',
+    url: `http://${PYTHON_APP || 'localhost'}:5000/gtfs-feed-info`,
```
kq-li: probably best if we don't hardcode this port

src/API.js
```diff
@@ -18,6 +18,7 @@ class API extends ApplicationAPI {
       Routers.AlertsRouter,
       Routers.AllStopsRouter,
       Routers.DelayRouter,
+      Routers.FeedInfoRouter,
```
alanna-zhou: should probably name this GTFSFeedInfo - not clear which feed this is, because we also have a live-tracking feed!

src/routers/v1/FeedInfoRouter.js
```diff
@@ -0,0 +1,20 @@
...
+import GTFSUtils from '../../utils/GTFSUtils';
+import ApplicationRouter from '../../appdev/ApplicationRouter';
+
+class FeedInfoRouter extends ApplicationRouter<Array<Object>> {
+  constructor() {
+    super(['GET']);
+  }
+
+  getPath(): string {
+    return '/feedinfo/';
```
alanna-zhou: we've been using camel case (ex. /allStops) -- could this be feedInfo instead?

Issue #261, opened 2019-09-25
Walking route has strange start and end times 


alanna-zhou:
![image](https://user-images.githubusercontent.com/13739525/65644407-2772df80-dfc2-11e9-917c-1a83db870b92.png)

PR #275, opened 2019-11-10
Fix "promise chaining cycle" due to modifying promises in walking route

alanna-zhou:
# UPDATE: looks like this change doesn't actually fix anything -- I'm going to leave this PR open as I'm figuring this out and add stuff as I go :'(

## Overview
So right now, my walking route time fix is extremely flakey because there's this promise cycle detection error, and I finally figured out what the issue is. And it's because I've been directly modifying the `directions` array that's a promise!

## Changes Made
So instead of directly modifying the `directions`'s first element's `startTime` and `endTime`, I created a new directions and appended the rest of the existing information so that I don't mess with all of this asynchronous jazz.

I've also renamed the return types to `Promise`s because this should be fixed eventually.

## Test Coverage
Currently deployed on dev and doesn't break anything. I'll let it sit for a while and then merge this PR in. 

## Next Steps
I think I will write a Python script that checks for a valid `ithaca-transit-backend` JSON response -- because I think that would have been super helpful for me to use than to manually cross check with our documentation. Or I think I might talk to someone about adding this to `integration`.

## Related PRs or Issues 
#274 This was the culprit PR that had the promise cycle detection issue.
#261 Actually addresses this issue finally.

### DIFF
--- a/src/utils/ParseRouteUtils.js
+++ b/src/utils/ParseRouteUtils.js
@@ -467,8 +467,21 @@ function parseRoutes(
       const numberOfLegs = legs.length;
 
       // string 2018-02-21T17:27:00Z
-      const { departureTime } = legs[0];
-      const arriveTime = legs[numberOfLegs - 1].arrivalTime;
+      let { departureTime } = legs[0];
+      let arriveTime = legs[numberOfLegs - 1].arrivalTime;
+
+      // Readjust the walking start and end times by accounting for the buffer
+      // times that were initially passed into Graphhopper to get routes
+      if (busRoute.transfers === -1) {
+        let startTimeMs = originalDepartureTimeMs;
+        let endTimeMs = originalDepartureTimeMs + busRoute.time;
+        if (isArriveByQuery) {
+          startTimeMs = originalDepartureTimeMs - busRoute.time;
+          endTimeMs = originalDepartureTimeMs;
+        }
+        departureTime = convertMillisecondsToISOString(startTimeMs);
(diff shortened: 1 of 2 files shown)


src/utils/ParseRouteUtils.js
```diff
@@ -676,14 +676,17 @@ function parseRoutes(
         const walkDepartureTime = convertMillisecondsToISOString(startTimeMs);
         const walkArrivalTime = convertMillisecondsToISOString(endTimeMs);
 
-        directions[0].startTime = walkDepartureTime;
-        directions[0].endTime = walkArrivalTime;
+        const walkDirections = {
+          ...directions[0],
```
Omarrasheed: out of curiosity what does the `...direction[0]` do/mean

alanna-zhou: directions is an array and walking will always have 1 direction in directions

Omarrasheed: lmao i was referring to the `...` but Dr Chan explained it to me so we chillin. tyty

src/utils/ParseRouteUtils.js
```diff
@@ -676,14 +676,17 @@ function parseRoutes(
...
+          startTime: walkDepartureTime,
+          endTime: walkArrivalTime,
```
young-k: nifty use of the `...` syntax here, I love it

kq-li: creds to me :^)

PR #278, opened 2019-12-02
Add notification endpoints for iOS

Omarrasheed:
## Overview

Add endpoints for ios to hit. We don't like to block ppl here

## Changes Made

Make endpoints that return true if all fields r there

## Test Coverage

Builds

## Next Steps

Actually implement the stuff once I finish distributed


src/routers/v2/DelayNotification.js
```diff
@@ -0,0 +1,38 @@
...
+      || typeof req.body.deviceToken !== 'string'
+      || !req.body.tripID
+      || typeof req.body.tripID !== 'string'
+      || !req.body.stopID
+      || typeof req.body.stopID !== 'string'
+      || !req.body.uid
+      || typeof req.body.uid !== 'string') {
+      return false;
+    }
+    return NotificationUtils.notifyForDelays(
```
kevinchan159: you can do `{ deviceToken, tripID, uid } = req.body`

src/routers/v2/DepartNotification.js
```diff
@@ -0,0 +1,41 @@
...
+      || typeof req.body.tripID !== 'string'
+      || !req.body.stopID
+      || typeof req.body.stopID !== 'string'
+      || !req.body.uid
+      || typeof req.body.uid !== 'string'
+      || !req.body.minutesBefore
+      || typeof req.body.minutesBefore !== 'number') {
+      return false;
+    }
+    return NotificationUtils.notifyForDeparture(
```
kevinchan159: same here

src/routers/v2/DepartNotification.js
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  getPath(): string {
+    return '/departNotification/';
+  }
+
+  // eslint-disable-next-line require-await
+  async content(req: Request): Promise<any> {
+    if (
+      !req.body
```
meganle: Don't need to change it in this PR but we should look into using a library for validating so we avoid huge if statements next semester @alanna-zhou . I would look into https://www.npmjs.com/package/validate or https://validatejs.org/

Issue #284, opened 2020-01-29
Routes that have start times before the Leave now time are showing up


alanna-zhou:
![image](https://user-images.githubusercontent.com/13739525/73404724-2d563580-42c0-11ea-8e4c-0c784b8ac439.png)

PR #285, opened 2020-02-05
Remove bus routes with an earlier start time than departure time

alanna-zhou:
## Overview
Right now, we show routes that are some 20-30 minutes before the departure time, be it arrive by or not.

I still need to investigate why this is happening, but for now I've put in a trivial fix so that our app doesn't look weird.

image

## Changes Made
Trash the bus route if it's before the time that the user expects the leave when we parse bus routes.

## Test Coverage
Tested locally and on the dev server under `transit-dev:alanna`.

## Next Steps
Will actually investigate what's going on and fix the early routes. 
Also, an issue with this change is that if there's some walking time before the bus route, it may still be earlier than the departure time. 

## Related PRs or Issues
#284

### DIFF
--- a/src/utils/ParseRouteUtils.js
+++ b/src/utils/ParseRouteUtils.js
@@ -183,6 +183,12 @@ async function condenseRoute(
       if (startTime < departureTimeNowMs - ONE_MIN_IN_MS - delayMs) {
         return null;
       }
+
+      // Temporary fix to prevent bus routes starting before departure time from
+      // being returned.
+      if (startTime < departureTimeNowMs) {
+        return null;
+      }
     }
 
     if (previousDirection
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,7 +13,7 @@ services:
       - /usr/src/app/node_modules
 
   ghopper:
-    image: cornellappdev/transit-ghopper:v1.1.6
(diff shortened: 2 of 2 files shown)


src/utils/ParseRouteUtils.js
```diff
@@ -183,6 +183,12 @@ async function condenseRoute(
       if (startTime < departureTimeNowMs - ONE_MIN_IN_MS - delayMs) {
         return null;
       }
```
kevinchan159: curious what is the value of `delayMs` here?

alanna-zhou: i think it's literally the delay, so that we make sure that users can still catch buses that are late. but i feel like delay should actually be added onto departureTimeNow ?

alanna-zhou: agh ive been comparing between tranist prod and transit beta (which is running my change), and dev just seems to make a lot more sense so idk

kevinchan159: yeah I feel like the issue we should be looking into is how these routes make it past this check. I would imagine the delayMs isnt that large?

young-k: @kevinchan159 if i remember correctly from when I checked the blame, you introduced delayMs into the codebase!

PR #292, opened 2020-04-05
Refactor `/tracking` to make requests to the microservice and return valid vehicle info

alanna-zhou:
Again, sorry if I added too many reviewers, everyone goes MIA on me so I wanted to spam as many people as possible. Please help me get this approved ASAP so Android can ship within this semester!!

## Overview
Currently, `/tracking` is an endpoint for `ithaca-transit-backend`, and right now it isn't returning data for the buses, and so the iOS and Android clients can't draw the buses on the map. This is the second step in my refactoring plan for addressing this issue:

1. Microservice parses protobuf for vehicle information from the TCAT third-party endpoint on vehicle positions
2. `/tracking` on ithaca-transit-backend needs to be modified so that it uses the newly merged ithaca-transit-microservice's `/vehicles` which takes in a `tripID` and `routeID` and returns important info like `longitude` and `latitude` which will help draw the buses. I originally wanted to make a new endpoint `/vehicle` but I decided that the old URL we were hitting for the data was completely irrelevant and no longer updated by TCAT's third-party vendor, so I ended up just scratching all of that and rewrote `/tracking`
3. We need to update the swagger docs that shows the updated `/tracking`, which is now able to take in an array of info and return the info as an array: `[{tripID, routeID}, ..., {tripID, routeID}]` returned as `[{bearing, longitude, latitude, speed, etc}, {bearing, etc}, ... {}]`

This PR is step 2, I will be making one more PR soon on `ithaca-transit-backend` that addresses step 3.

## Changes Made
I rewrote `/tracking` by adding new `fetchVehicles` and `getVehicleInformation` functions, which essentially mirrors what `fetchRTF` and `getDelayInformation` are doing. It made sense to do this because it was pretty clean and easy to understand already.

In general, the OG code for `/tracking` was just absolutely h o r r e n d o u s as I could not understand what the heck was going on! Apologies that this hard work was completely scrapped!!

I also had to update the `docker-compose.yml` so that the most recent images with the most recent GTFS data could be used while I was testing locally. Most notably, the image `transit-python:alanna` is my current test Docker image that has the step 1 change in master. 

## Test Coverage
Works locally. Will be letting this sit on the dev server for a bit though!

## Next Steps
Updating swagger docs.

## Related PRs or Issues
Step 1: https://github.com/cuappdev/ithaca-transit-tcat-microservice/pull/49

## Screenshots 
Screen Shot 2020-04-05 at 12 35 39 PM

### DIFF
--- a/src/utils/RealtimeFeedUtils.js
+++ b/src/utils/RealtimeFeedUtils.js
@@ -36,21 +36,19 @@ async function fetchVehicles(): Object {
  */
 async function getTrackingResponse(requestData: Object): Object {
   LogUtils.log({ message: 'getTrackingResponse: entering function' });
-  const trackingInformation = [];
   const vehicles = await fetchVehicles();
+  console.log('vehicles', vehicles);
 
-  // for each input
-  await Promise.all(requestData.map((data) => {
+  const trackingInformation = requestData.map((data) => {
     const { routeID, tripID } = data;
     const vehicleData = getVehicleInformation(routeID, tripID, vehicles);
-
     if (!vehicleData) {
       LogUtils.log({ message: 'getVehicleResponse: noData', vehicleData });
-      return false;
+      return null;
     }
-    trackingInformation.push(vehicleData);
-    return true;
-  }));
+    return vehicleData;
+  }).filter(Boolean);
+
   return trackingInformation;
 }
 
@@ -61,7 +59,11 @@ async function getTrackingResponse(requestData: Object): Object {
  * @param rtf
  * @returns Object
  */
-function getDelayInformation(stopID: String, tripID: String, rtf: Object): ?Object {
+function getDelayInformation(
+  stopID: ?String,
+  tripID: ?String,
+  rtf: ?Object,
+): ?Object {
   // rtf param ensures the realtimeFeed doesn't update in the middle of execution
   // if invalid params or the trip is inactive
   if (!stopID
@@ -88,10 +90,10 @@ function getDelayInformation(stopID: String, tripID: String, rtf: Object): ?Obje
 }
 
 function getVehicleInformation(
-  routeID: String,
-  tripID: String,
-  vehicles: Object,
-): ?Array<Object> {
+  routeID: ?String,
+  tripID: ?String,
+  vehicles: ?Object,
+): ?Object {
   // vehicles param ensures the vehicle tracking information doesn't update in
   // the middle of execution
   if (!routeID
@@ -105,9 +107,8 @@ function getVehicleInformation(
     });
     return null;
   }
-  return Object.values(vehicles).filter(
-    v => v.routeID === routeID && v.tripID === tripID,
-  )[0];
+
+  return Object.values(vehicles).find(v => v.routeID === routeID && v.tripID === tripID);
 }
 
 export default {
--- a/src/utils/RealtimeFeedUtils.js
+++ b/src/utils/RealtimeFeedUtils.js
@@ -14,13 +14,11 @@ async function fetchRTF(): Object {
 }
 
 async function fetchVehicles(): Object {
-  console.log('fetching vehicles');
   const options = {
     ...Constants.GET_OPTIONS,
     url: `http://${PYTHON_APP || 'localhost'}:5000/vehicles`,
   };
   const data = await RequestUtils.createRequest(options, 'Vehicles request failed');
-  console.log('data:', data);
   return JSON.parse(data);
 }
 
@@ -93,7 +91,7 @@ function getVehicleInformation(
   routeID: String,
   tripID: String,
   vehicles: Object,
-): ?Object {
+): ?Array<Object> {
   // vehicles param ensures the vehicle tracking information doesn't update in
   // the middle of execution
   if (!routeID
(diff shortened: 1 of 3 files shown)


src/utils/RealtimeFeedUtils.js
```diff
@@ -3,140 +3,55 @@ import { PYTHON_APP } from './EnvUtils';
...
-          name: busInfo.Name,
-          opStatus: busInfo.OpStatus,
-          routeID: busInfo.RouteId,
-          runID: busInfo.RunId,
-          speed: busInfo.Speed,
-          tripID: busInfo.TripId,
-          vehicleID: busInfo.VehicleId,
-        };
-      })(busFound);
+  await Promise.all(requestData.map((data) => {
```
kevinchan159, line 43: you don't need `Promise.all` here since you don't perform any async operations right?

kevinchan159, line 43: also might be cleaner if you do:
```
const trackingInformation = requestData.map((data) => {
...
  if (!vehicleData) return null
  return vehicleData
}).filter(Boolean)
```

src/utils/RealtimeFeedUtils.js
```diff
@@ -172,8 +87,33 @@ function getDelayInformation(stopID: String, tripID: String, rtf: Object): ?Obje
...
+    LogUtils.log({
+      category: 'getVehicleInformation NULL',
+      routeID,
+      tripID,
+    });
+    return null;
+  }
+  return Object.values(vehicles).filter(
+    v => v.routeID === routeID && v.tripID === tripID,
+  )[0];
```
kevinchan159, line 110: this would error if the return of `filter` is an empty array right? we should prob check the length before returning `[0]`

src/utils/RealtimeFeedUtils.js
```diff
@@ -172,8 +87,33 @@ function getDelayInformation(stopID: String, tripID: String, rtf: Object): ?Obje
   };
 }
 
+function getVehicleInformation(
+  routeID: String,
+  tripID: String,
+  vehicles: Object,
+): ?Array<Object> {
```
kevinchan159, line 94: also not sure how the data is formatted but should the return type here by `?Array<Object>` or `?Object`?

alanna-zhou: ohh u right

src/utils/RealtimeFeedUtils.js
young-k: WOW

src/utils/RealtimeFeedUtils.js
```diff
@@ -172,8 +87,33 @@ function getDelayInformation(stopID: String, tripID: String, rtf: Object): ?Obje
...
+  // vehicles param ensures the vehicle tracking information doesn't update in
+  // the middle of execution
+  if (!routeID
+    || !tripID
+    || !vehicles
+    || vehicles === {}) {
```
young-k, line 100: would `vehicles === {}` not be caught by `!vehicles`?

src/utils/RealtimeFeedUtils.js
```diff
@@ -172,8 +87,33 @@ function getDelayInformation(stopID: String, tripID: String, rtf: Object): ?Obje
   };
 }
 
+function getVehicleInformation(
+  routeID: String,
+  tripID: String,
```
young-k, line 92: if these are all nullable, shouldn't they be marked as `?`, or is the `!routeID` check to see if the string is non-empty?

src/utils/RealtimeFeedUtils.js
```diff
@@ -172,8 +87,33 @@ function getDelayInformation(stopID: String, tripID: String, rtf: Object): ?Obje
...
+    || vehicles === {}) {
+    LogUtils.log({
+      category: 'getVehicleInformation NULL',
+      routeID,
+      tripID,
+    });
+    return null;
+  }
+  return Object.values(vehicles).filter(
+    v => v.routeID === routeID && v.tripID === tripID,
```
young-k, line 109: hm, isn't it convention to have `v =>` on the previous line to make this easier to read?

alanna-zhou: ohh i didnt know that, gotcha!

src/utils/RealtimeFeedUtils.js
alanna-zhou: oh... i did not know that `?` meant they are nullable.. i shall go do that now thanks

src/utils/RealtimeFeedUtils.js
alanna-zhou: for some reason this is what `getDelayInformation` is checking with the `!rtf` and `rtf === {}` so i just copied it over... but what you're saying makes more sense to me so i'll go do some investigation

kevinchan159: it does not catch it im pretty sure so you need the check

src/utils/RealtimeFeedUtils.js
alanna-zhou: u right.. i shall defer to your way :) thx

src/utils/RealtimeFeedUtils.js
alanna-zhou: yes u r right

alanna-zhou: omg... legend

PR #295, opened 2020-04-11
Update `routeID` to be `routeNumber` in `\tracking` for consistency

alanna-zhou:
## Overview
Right now `/tracking` requires `routeID` and returns `routeID` but the Android team helped me realize that our backend views the `routeID` in the GTFS data (string in the GTFS) but returns it as a `routeNumber` (type Number). So I needed to make that change so that backend is consistent. I didn't choose to modify this in the microservice, because the whole point of the python code is to abstract that away from having to deal with what the clients know about our backend. 

## Changes Made
Changed references of `routeID` to `routeNumber` and changed data type of string to number.
Updated docker compose to run locally.
Updated swagger docs to reflect new required field in request body and new field in response body.

## Test Coverage
Works locally, will be pushing onto servers soon.

## Related PRs or Issues 
#293 updated

## Screenshots
Screen Shot 2020-04-11 at 2 08 10 PM

### DIFF
--- a/src/utils/RealtimeFeedUtils.js
+++ b/src/utils/RealtimeFeedUtils.js
@@ -107,7 +107,7 @@ function getVehicleInformation(
     return null;
   }
 
-  const vehicleData = Object.values(vehicles).find(v => v.routeID == routeNumber && v.tripID === tripID);
+  const vehicleData = Object.values(vehicles).find(v => parseInt(v.routeID) === routeNumber && v.tripID === tripID);
   if (!vehicleData) {
     LogUtils.log({
       category: 'getVehicleInformation no data',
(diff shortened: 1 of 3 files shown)


src/utils/RealtimeFeedUtils.js
```diff
@@ -90,25 +89,43 @@ function getDelayInformation(
...
       category: 'getVehicleInformation NULL',
-      routeID,
+      routeNumber,
       tripID,
     });
     return null;
   }
 
-  return Object.values(vehicles).find(v => v.routeID === routeID && v.tripID === tripID);
+  const vehicleData = Object.values(vehicles).find(v => parseInt(v.routeID) === routeNumber && v.tripID === tripID);
```
kevinchan159, line 110: maybe split this into multiple lines

PR #302, opened 2020-08-28
Update PR template

alanna-zhou:
## Overview
This is super meta but I getting really annoyed always having to delete the intro on how to make a good PR title. I think everyone here already knows how to do this...

Basically everytime I delete it, Github detects the markdown as the start of an issue:

Screen Shot 2020-08-28 at 12 02 20 PM

## Changes Made
* Remove redundant descriptions (like who doesn't understand this stuff after reading the titles??)
* Add an HTML tag template for adding images so that we can resize and that is *actually* useful instead of having to click on a dropdown and seeing some image swallow up your whole screen

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,0 +1,17 @@
+# Contributing to Ithaca Transit Backend
+👍🎉 First off, congrats on getting put on this pod 😂🎉👍
+
+The following is a set of guidelines for contributing to our backend. These are mostly guidelines, not rules. Use your best judgment, and feel free to propose changes to this document in a pull request.
+
+# Making PRs
+We want our PRs to be concise but informative. Some pointers as per our PR template:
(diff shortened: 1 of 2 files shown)


whoiswillma:
> reminders never hurt and will help new contributors

I think the point is that the reminders are a paper cut, and so do hurt, even if just a little bit.

The new maintainers thing is a good point. I think adding a CONTRIBUTING.md could solve that issue.

aliciaxw:
> > reminders never hurt and will help new contributors
> 
> I think the point is that the reminders are a paper cut, and so do hurt, even if just a little bit.
> 
> The new maintainers thing is a good point. I think adding a CONTRIBUTING.md could solve that issue.

sure, having that file would do the job

alanna-zhou:
thanks for the comments guys, bump on this @aliciaxw @whoiswillma 

made a CONTRIBUTING.md! @whoiswillma does it suffice to put this new doc in `.github`?

PR #308, opened 2020-10-11
Begin v3/route by copying v2 into a new v3

alanna-zhou:
## Overview
Some really redundant/jank things we want to get rid of from v2:
* routeSummary
* delays
* routeID -- no longer using for analytics!
* routeNumber -- since we delete routeID, we can change this to routeID

## Changes Made
* Copy v2 into a v3 route

## Test Coverage
Works locally, and although it's a minor change I'll be deploying to the server

## Next Steps
* Might want to consider refactoring delays to return the delays for a specific stop

## Related PRs or Issues
#307

### DIFF
--- a/src/routers/v3/RouteRouter.js
+++ b/src/routers/v3/RouteRouter.js
@@ -1,0 +1,61 @@
+// @flow
+import type Request from 'express';
+import AnalyticsUtils from '../../utils/AnalyticsUtils';
+import ApplicationRouter from '../../appdev/ApplicationRouter';
+import LogUtils from '../../utils/LogUtils';
+import RouteUtils from '../../utils/RouteUtils';
+
+class RouteRouter extends ApplicationRouter<Object> {
+  constructor() {
+    super(['POST']);
+  }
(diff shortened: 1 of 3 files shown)


src/routers/v3/RouteRouter.js
```diff
@@ -0,0 +1,61 @@
...
+      destinationName,
+      end,
+      arriveBy,
+      originName,
+      start,
+      time,
+      uid,
+    } = req.body;
+
+    const isArriveBy = (arriveBy === '1' || arriveBy === true || arriveBy === 'true' || arriveBy === 'True');
```
aliciaxw, line 28: why are there so many possible inputs for `arriveBy`?

aliciaxw (approved):
nice

src/routers/v3/RouteRouter.js
alanna-zhou: that's a good question :) it's because people send in all types of random stuff to us so this is a precaution

PR #312, opened 2020-11-05
Add v3/tracking and update v3/delays, v3/route, and swagger docs

alanna-zhou:
## Overview
We have a lot of naming issues and today I decided to fix all the wacko naming and types to match as close as possible to [General Transit Feed Specification (GTFS)](https://gtfs.org/). 🎊 🎉 

**If you are on iOS/Android looking at this PR, make sure you notice the subtle capitalization differences.** If you ever have questions about what any of these fields mean in the context of buses, I've deliberately matched up the naming with GTFS standard, with one difference being the first letter is lowercase for our backend (Ex. `routeId` vs GTFS' `RouteId`). See the bottom of this long description for helpful request examples! 

### Before
* **`routeId`**: our own unique hashed id for analytics (`str`)
* **`routeNumber`** (`int` _sometimes_): GTFS `routeId` (`str`)
* **`tripIdentifiers`** (`[str]`): array of GTFS `tripId`s (`[str]`)
* **`stopID`** (`int`): GTFS `stopId` (`string`)

### Now
* **`routeId`** (`str`): GTFS `routeId` (`str`)
* **`tripIds`** (`[str]`): array of GTFS `tripId`s (`[str]`)
* **`stopId`** (`str`): GTFS `stopId` (`str`)

## Changes Made
* Update `docker-compose.yml`'s graphhopper version so the code can be run locally with the latest GTFS data
* Remove the old formatting data for v3/delays since we can just rely on v2 to be backwards compatible
* Update all swagger docs including old typos and types
* Our routers were getting weird overlaps (hard to explain but basically we were routing multiple versions of delays or tracking to v2)
* Get rid of analytics, we weren't using it away and it wasn't helpful data (the `routeId` that we were using to cache a particular route wasn't actually unique at all so it wasn't like we were able to even store routes that users clicked on properly)
* Ensure `tripIds` returns `[]` instead of `null`
* Slim down `v3/tracking` as per iOS's v3 requests

## Test Coverage
Works locally, will be letting this sit on the dev server for a bit before deploying it to production

## Next Steps
I'm going to make a separate PR that fixes all of the documentation -- a bit too overwhelming if I also included it in this PR, so don't comment on the documentation, just focus on the code @ reviewers

## Related PRs or Issues
Closes #295 

## v3/route

**Example Request Body**:
```{
  "end" : "42.43564391129678,-76.51053786277771",
  "uid" : "B1CD4726-0CA8-41DC-A2F4-01D5D8F3A3E0",
  "time" : 1604536485.363683,
  "destinationName" : "Wegmans",
  "start" : "42.450884,-76.4936187",
  "arriveBy" : false,
  "originName" : "Auden Ithaca"
}
```
**Example Response Body**: [response.txt](https://github.com/cuappdev/ithaca-transit-backend/files/5491630/response.txt)

## v3/delays
<details><summary>Example Request in Postman</summary>
<p>
Screen Shot 2020-11-04 at 6 12 55 PM
</p>
</details>

## v3/tracking
<details><summary>Example Request in Postman</summary>
<p>
Screen Shot 2020-11-04 at 6 05 20 PM
</p>
</details>


src/utils/RealtimeFeedUtilsV3.js
```diff
@@ -0,0 +1,150 @@
...
+ * routeID : String,
+ * tripID : String
+ * },... ]
+ * NOTE: Because we need to provide backwards compatibility with old iOS clients
+ * we have to follow their janky way of routeID input is String but routeID
+ * output is Number. This "routeID" is also named jankily, which is supposed to
+ * be routeNumber from v2/route/. We cast this to Number in getVehicleInformation.
+ *
+ *
+ */
```
aliciaxw, line 39: marking this for reference when you update your documentation in a later pr :-)

aliciaxw (approved):
looks great, and good move changing naming to be consistent with GTFS!

src/routers/v3/RouteRouter.js
```diff
@@ -45,15 +44,12 @@ class RouteRouter extends ApplicationRouter<Object> {
         destinationName,
         end: routes[0].endCoords,
         originName,
-        routeId: routes[0].routeId,
```
raahimenon, line 48: do they not need routeId/routeName/whatever it is? I thought this was being converted to the GTFS routeId

src/utils/RealtimeFeedUtilsV3.js
```diff
@@ -0,0 +1,150 @@
...
+  const data = await RequestUtils.createRequest(options, 'Vehicles request failed');
+  return JSON.parse(data);
+}
+
+/**
+ * Given an array of { routeID, tripID },
+ * Return bus information
+ * Input:[
+ * {
+ * routeID : String,
```
raahimenon, line 30: is this routeID capitalized because that's what frontend stores?
why are we being inconsistent with their naming?

src/routers/v3/RouteRouter.js
alanna-zhou: no this wasn't a value in the first place -- a route id corresponds to each direction, but this is an entire route which doesnt have a route id -- i know, confusing

src/utils/RealtimeFeedUtilsV3.js
alanna-zhou: i used to use routeID last semester because i didn't think naming mattered until i realized that there were inconsistencies everywhere!!!

like i said, i'm leaving updating the documentation to another PR, too overwhelming if i also updated documentation in this one

src/routers/v3/RouteRouter.js
alanna-zhou: if you read my description you will notice that this is our own routeId, which is from our analytics code -- has nothing to do with an actual GTFS routeId, which corresponds to a direction

PR #313, opened 2020-11-06
[2/2] Fix v3 naming schemes & add back dummy endpoints to ensure back compat

alanna-zhou:
## Overview
Fixes up loose ends on #312 because I deleted some endpoints -- but turns out I think iOS expects to still hit them, so I made them dummy endpoints instead (just return `null`). Made me a bit sad because this also meant the return of `AnalyticsUtils` (we still need it to add a dummy `routeId` field to `v1/route` and `v2/route` endpoints) but at least I got rid of the cache so this frees up some space on the server at least.

## Changes Made
* Update documentation 
* Make `AnalyticsUtils` just a dummy `routeId` assign-er to v1 and v2 endpoints
* Add back `/routeSelected` for v1 and v2 but keep it a dummy endpoint that just returns `null`

## Test Coverage
Tested on dev server, it works :)
Note: for whoever future backend dev looking at this, if you want to revert this commit, you also need to revert #312 because this PR is what fixes some back compat issues there

## Next Steps
Make some tests on integration to assert fields exist, and at the very least assert that they have some null value

## Related PRs or Issues
Part 2 of #312


yuna878 (approved):
Looks good! The null endpoint is kinda wack tho lol

PR #314, opened 2020-12-02
Add Delay Route for Backwards Compatibility with iOS

raahimenon:
## Overview
iOS uses a `v2/delay` route instead of `v2/delays` - for the sake of backwards compatibility with people who don't update, I added a delay route. 

## Changes Made
Added a new `DelayRouter` with similar functionality as `DelaysRouter`, but it only accepts a single trip, and accepts it via the query parameters instead of the request body (as in `v1`).  

## Test Coverage
Some manual testing, and ran integration locally.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,7 +13,7 @@ services:
       - /usr/src/app/node_modules
 
   ghopper:
-    image: cornellappdev/transit-ghopper:v1.2.9
+    image: cornellappdev/transit-ghopper:11_29_20
     ports:
       - "8988:8988"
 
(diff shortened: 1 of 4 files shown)


alanna-zhou (approved):
looks great :)) make sure that before you merge this into master, that you build a docker image on this branch called `transit-dev:raahi` and test that by deploying onto `transit-dev` first -- and once that is fine for a day, then merge into master, and build a docker image on master called `transit-node:<DATE>` and first deploy onto `transit-dev` to see all is well, and then deploy onto `transit-prod`

alanna-zhou (approved):
looks good :) 

we usually do a separate PR for updating the docker compose but i also get lazy and just include it in my latest PR haha

PR #326, opened 2022-01-23
update ghopper wget url to modern aws path name

raahimenon:
## Overview
Ghopper GTFS wget URL was using old S3 path conventions that are planned to be deprecated. 

## Changes Made
Updated URL to modern style.

## Test Coverage
Built Dockerfile and it worked.


chalo2000 (approved):
@raahimenon I'll let you merge this in

PR #341, opened 2024-06-11
fixed logic for arriveBy

cindy-x-liang:
## Overview
previously arrive by worked the same as leave at
I changed it so that most of the routes returned are before the arrive by time
we don't implement the arrive by logic, graphhopper does, so it is slightly buggy but I'm not sure if we can fix that

image
image

^^ this is what it looks like when i deploy on the dev server


Aayush-Agnihotri (approved):
Looks good to me! I wonder if we can just have frontend filter the routes that don't make sense but I think its good for now.

PR #343, opened 2024-10-23
Nicole delay final time

nfq2:
## Overview
Fix bug where the final delay time is not accurate. 

## Changes Made
Added logic in `RouteUtilsV3.js` to iterate over each route in finalBusRoutes, calculates total delay from bus segments, and apply this delay to the 'walk' segment if it follows a delayed bus segment.

## Test Coverage
Tested routes with no delays, delays, final destination at bus stop, final destination not at bus stop. Finding routes with multiple different bus and walking segments that had delays were a bit difficult, so I just created my own route object with directions to test the code. 

## Related PRs or Issues
[https://github.com/cuappdev/ithaca-transit-ios/issues/392](url)

## Screenshots
Original bug (delay not reflected in final time): 
image

Request body:
{
  "end": "42.4403, -76.4962",
  "uid": "B1CD4726-0CA8-41DC-A2F4-01D5D8F3A3E0",
  "time": 1729604959,
  "destinationName": "Thompson and Bleeker",
  "start": "42.4852, -76.4909",
  "arriveBy": false,
  "originName": "Target"
}

One of the routes contains a bus delay: 
image

Delay time reflected in final walking segment: 
image

### DIFF
--- a/src/utils/RouteUtilsV3.js
+++ b/src/utils/RouteUtilsV3.js
@@ -254,6 +254,30 @@ async function getSectionedRoutes(
     isArriveBy,
   );
 
+  LogUtils.log('new testing delay 1');
+  LogUtils.log(finalBusRoutes.length);
+
+  // finalBusRoutes.forEach((route) => {
+  //   let totalDelay = 0;
+  //   const { directions } = route;
+  //   LogUtils.log('testing delay');
+  //   for (let i = 0; i < directions.length; i++) {
+  //     const segment = directions[i];
+  //     const { delay } = segment;
+  //     LogUtils.log('heres the delay');
+  //     LogUtils.log(delay);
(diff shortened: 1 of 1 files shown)


src/utils/RouteUtilsV3.js
cindy-x-liang, line 1: Does this consider the delay for the first walking portion before you get onto the bus? I think you will also need to check if the first bus in the directions is delayed and if so, add that time to the first walking portion of the directions (when you walk to the bus stop).

Aayush-Agnihotri: ^

cindy-x-liang (approved):

Aayush-Agnihotri (approved):

PR #346, opened 2025-02-20
Fix Swagger

nfq2:
## Overview
Edited Swagger doc example request bodies for /delays and /tracking since it was causing confusion with frontend resulting in a delay bug. 

## Changes Made
swagger.json

## Test Coverage

Before fix: 
image
image

After fix: 
image
image


Aayush-Agnihotri (approved):
lgtm

PR #353, opened 2025-03-26
Add Route Reporting Functionality

daisychangm:
## Overview

I set up a SQLite database to store route reports, which records the following data for each report: vehicle ID, congestion level, device token, and an optional time stamp. I also added endpoints to insert a new report and fetch all reports for a bus.

## Changes Made

I defined two endpoints in `RouteReportingController.js` to fetch and insert reports, and I implemented the logic for those endpoints in `fetchReportByBus` and `insertReport` in `RouteReportingUtils.js`. I also updated the run command in `package.json` to prevent the server from restarting everytime vehicle data is fetched.

## Test Coverage

I manually tested this feature with Postman. A picture of a sample insert request and response is shown below.

## Screenshots

image


nfq2 (commented):
We should probably get rid of the pycache files?

daisychangm:
Okay, I removed them

nfq2 (approved):
lgtm

Aayush-Agnihotri (approved):
lgtm

PR #354, opened 2025-03-26
Web scraped restaurants from API and added them to database

lauren1703:
## Overview

I found an online API and web scraped all the restaurants in Ithaca. I added a table in the database to store the restaurants and created an endpoint to retrieve this data.

## Changes Made

I created the `restaurants.py` for the web scraping function. I also created a table in SQLite in `models.py`, defined the `insert_restaurant()` function in `database.py` and added to the populate_db() function. I defined the endpoint in `EcosystemController.js` and updated `EcosystemUtils.js` accordingly.

## Test Coverage

I manually tested using Postman and made sure all the restaurants were returned with the correct info.

## Screenshots

image


nfq2 (commented):
you should also update the swagger file for ur new endpoint and add versioning to the api endpoint (the route should be: /api/v1/restaurants)

everything else lgtm

PR #355, opened 2025-04-05
indicate nullable field on swagger doc

nfq2:
Indicate when fields are possibly null for frontend decoding errors. Also updated the models to be compatible with OpenAPI 3.0.


daisychangm (approved):
LGTM

PR #358, opened 2025-08-29
Get RTF Data from source in Notification Loop

MrPeterss:
### Description

Initially, `rtfData` was only being declared and never initialized, causing no notifications to be sent at all, even though the endpoint works.


Aayush-Agnihotri (approved):
lgtm

PR #365, opened 2026-03-04
Event Form - Wyatt

WyattC-ctrl:
Updated Chimdi Event form to add images as well -- See documentation below

## Overview

In this pull request, I introduce in-house event form submission and moderation functionality to Navi, creating the backend logic needed to enable users to submit requests to host events directly through the app. These event requests are persisted in the backend on the SQLite database, and moderated by an approval workflow. The goal is to replace the need for external tools (e.g. Google Forms) with a fully integrated backend solution, and to facilitate real-time updates to both submitters and administrators.

## Changes Made

**Change 1: Introduce event form persistence via database migrations**
I added a new migration to introduce an event_forms table, which stores event submission metadata including:
- submitting user (`netid`)
- event name and type (temporary tabling vs permanent hotspot)
- conditional fields for temporary events (dates and hosting organization)
- event description, location, and approval status
- images

**Change 2: Implement REST endpoints for event submission and retrieval**
I added REST endpoints and a corresponding controller to support the full event submission lifecycle:
- `POST /events/create-event` to submit a new event request
- `GET /events` to retrieve all submissions (admin-facing, includes public and private information)
- `GET /events/approved` to retrieve all approved events (public-facing, sanitized output)
- `PUT /events/:id` to approve or reject a submission (admin-only)
Swagger documentation was updated to reflect all new endpoints and request/response schemas.

I introduced a `toPublicEvent` helper to ensure that only public-safe fields are sent to non-admin clients. Internal identifiers and metadata (e.g. database IDs and timestamps) are excluded from public payloads, reducing frontend coupling to database schema.

**Change 4: Migration adjustments**
Combined all mitigations file into one

## Test Coverage

I manually tested the feature end-to-end using a combination of REST requests and WebSocket clients via multiple terminals.
- Verified that submitting an event via `POST /events/create-event` correctly inserts a record into the database
- Used `curl` to approve and reject events via `PUT /events/:id`, confirming that:

## Next Steps
- Confirmed Swagger documentation

### DIFF
--- a/EventFormsController.js
+++ b/EventFormsController.js
@@ -1,0 +1,143 @@
+import express from "express";
+import multer from "multer";
+import { S3Client, PutObjectCommand } from "@aws-sdk/client-s3";
+import { createEventForm, getAllEventForms, updateEventForm, getApprovedEventForms, toPublicEvent, eventsByNetid, getImageByFormId } from "../utils/EventFormsUtils.js";
+
+const router = express.Router();
+const upload = multer({ storage: multer.memoryStorage() });
+
+// 1. Initialize the S3 Client for DigitalOcean
+const s3Client = new S3Client({
+    endpoint: "https://nyc3.digitaloceanspaces.com",
+    forcePathStyle: false, 
+    region: "us-east-1",
+    credentials: {
+        accessKeyId: process.env.FORM_S3_ACCESS_KEY, //MUST INCLUDE THIS FOR IT TO WORK (located in .env)
+        secretAccessKey: process.env.FORM_S3_SECRET_KEY 
+    }
+});
+
+// Create an event form
+router.post("/events/create-event", upload.single('image'), async (req, res) => {
+
+  try {
+    const { netid, name, eventType, startDate, endDate, organizationName, location, about } = req.body;
+    
+    //access the image from req.file
+    const file = req.file;
+    let fileUrl = null; 
+
+    //check if image was actually uploaded 
+    if (file) {
+      const params = {
+        Bucket: process.env.FORM_IMAGE_BUCKET_NAME,
+        Key: `events/${Date.now()}_${file.originalname}`,
+        Body: file.buffer,
+        ACL: "public-read",
+        ContentType: file.mimetype
+      };
+      // Upload to DigitalOcean
+      await s3Client.send(new PutObjectCommand(params));
+      fileUrl = `https://${process.env.FORM_IMAGE_BUCKET_NAME}.nyc3.digitaloceanspaces.com/${params.Key}`;
+      //EDIT to have the correct path for correct folder
+
+    }
+
+    const eventForm = await createEventForm({ netid, name, eventType, startDate, endDate, organizationName, location, about, image_url: fileUrl});
+
+    // Broadcast a notification to the admin room that the event form has been created
+    const io = req.app.get("io");
+    io.to("admin").emit("eventForm:new", {message: "Event request submitted", event: eventForm});
+
+    // Return the event form to the requesting client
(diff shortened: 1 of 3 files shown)


EventFormsController.js
```diff
@@ -0,0 +1,143 @@
...
+    res.status(200).json({ success: true, message: "All event requests retrieved successfully", data: eventForms.map(toPublicEvent)});
+  } catch (error) {
+    console.error("Error getting all event forms:", error.message);
+    res.status(400).json({ success: false, message: "Error getting all event requests", error: error.message });
+  }
+});
+
+// Update an event form
+// NOTE: Only admins can update event forms
+// NOTE: id is the event form's id, stored as the primary key in the database
```
JoshD94, line 73: Where is the check/auth to make sure the user is an admin?

EventFormsController.js
```diff
@@ -0,0 +1,143 @@
...
+        secretAccessKey: process.env.FORM_S3_SECRET_KEY 
+    }
+});
+
+// Create an event form
+router.post("/events/create-event", upload.single('image'), async (req, res) => {
+
+  try {
+    const { netid, name, eventType, startDate, endDate, organizationName, location, about } = req.body;
+
```
JoshD94, line 25: This implementation looks solid, but for the netid, allowing the user to just input this in the body is not secure. We should instead extract the user's netid from their login information, hence this must be a protected route if I am understanding the purpose of this event forms correctly.

EventFormsUtils.js
```diff
@@ -0,0 +1,301 @@
...
+  }
+
+
+/**
+ * Converts an event form to a public event.
+ * 
+ * @param {Object} eventForm - The event form to convert.
+ * @returns {Object} - The public event.
+ */
+function toPublicEvent({ name, netid, eventType, startDate, endDate, organizationName, about, location, image_url }) {
```
JoshD94, line 287: I think you need to map the snake_case that the DB returns to the camelCase you are using. Because in your migration files for example, event type is listed as event_type instead of eventType. So you need to add mapping like this:

```
function toPublicEvent(row) {
  return {
    name: row.name,
    netid: row.netid,
    eventType: row.event_type,
    startDate: row.start_date,
    endDate: row.end_date,
    organizationName: row.organization_name,
    about: row.about,
    location: row.location,
    imageUrl: row.image_url
  }
}
```

EventFormsUtils.js
```diff
@@ -0,0 +1,301 @@
...
+    db.all(query, (err, rows) => {
+      if (err) {
+        console.error(err.message);
+        return reject(err);
+      }
+      resolve(rows);
+    });
+
+    // Close the database
+    db.close((err) => {
```
JoshD94, line 205: Make sure your db.close() call is inside the db.all callback because this might run into some race condition issues. Because the query might still be running and suddenly you already close the connection to the db.

EventFormsController.js
```diff
@@ -0,0 +1,143 @@
...
+    const {netid} = req.params; // netid is found in the url path
+    const eventForms = await eventsByNetid(netid);
+    res.status(200).json({ success: true, message: "All events matching the netid retrieved successfully", data: eventForms.map(toPublicEvent)});
+  } catch (error) {
+    console.error("Error getting all events matching the netid:", error.message);
+    res.status(400).json({ success: false, message: "Error getting all approved event requests", error: error.message });
+  }
+});
+
+// Get image by event form ID
```
JoshD94, line 125: Do we need this route? The GET /events/ already has the image urls in the response. Unless we need this for testing

JoshD94 (changes requested):

PR #366, opened 2026-03-06
Josh health

JoshD94:
## Overview

Add new healthcheck endpoint for availability monitoring later on.

## Changes Made

- Created health controller which runs a simple DB query to ensure DB connection
- Created health endpoint and added to api
- Added endpoint to swagger docs


WyattC-ctrl (approved):
Looks Good.

